PR #3, opened 2026-05-25
fix: apply post-processing to streaming transcription path

MathurAditya724:
## Summary

- The WebSocket streaming path was bypassing LLM cleanup, dictionary replacements, and per-app format rules — sending raw transcription text straight to the client while the REST path applied all three
- History entries from streaming sessions were also missing `cleaned_text`, LLM metadata, token counts, and cost tracking

## Changes

- **Extract** post-processing into a shared `apps/server/src/lib/post-process.ts` module (LLM cleanup, dictionary replacements, context-aware formatting, cost calculation)
- **Refactor** `transcribe.ts` to use the shared module instead of inline logic
- **Apply** post-processing in `stream.ts` on the final text before sending it to the client
- **Add** a `context` WebSocket message type so the renderer can forward the frontmost app context (used for per-app format rule matching)
- **Update** streaming history inserts to populate all fields (`cleaned_text`, `llm_provider`, `llm_model`, `input_tokens`, `output_tokens`, `cost_usd`)
- **Widen** the `StreamCallbacks.onFinal` type to accept `Promise<void>` since the streaming handler is now async

### DIFF
--- a/apps/electron/src/main/paste.ts
+++ b/apps/electron/src/main/paste.ts
@@ -20,7 +20,6 @@ async function pasteWindows(): Promise<void> {
 }
 
 async function pasteLinux(): Promise<void> {
-  // Try xdotool first (X11), fall back to wtype (Wayland)
   try {
     await execAsync("xdotool key ctrl+v");
   } catch {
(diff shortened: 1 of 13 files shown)


MathurAditya724:
All CI checks green (lint, typecheck, and builds for Linux, macOS, Windows). Self-review completed — fixed 8 issues including a stale-closure bug on `useStreaming`, a race in the AudioWorklet init, missing `cleaned` message handling on the client, unsafe `ws.send` after close, stale recorder state, duplicate WAV encoding code, and a `destroy()` that permanently killed the Streamer on non-streaming models. Also added the shimmering text animation for pill state labels.

PR #25, opened 2026-05-25
feat: add Windows cloud-only support

### LINKED ISSUE #24: make freestyle compatible with cloud-only mode on Windows
Since freestyle is not compatible for Windows but there is a temporarily way to use cloud-only mode on Windows.

## How to run it on Windows right now

It's not officially supported — it's on the roadmap for v0.6. If you want to get it working today, here's what needs changing:

Skip local Whisper (use OpenAI cloud instead)
The build script for whisper.cpp will fail on Windows. You can work around this by only using the OpenAI
backend.

Fix the paste — in frontend/main/paste.ts, replace the osascript call with:
// Windows: use PowerShell to send Ctrl+V
await execAsync(powershell -command "Add-Type -AssemblyName System.Windows.Forms;   [System.Windows.Forms.SendKeys]::SendWait('^v')")

Fix the hotkey — in frontend/main/hotkey.ts, the Fn key name ('FN') may differ on Windows. You'd need to
change the hotkey to something like F13 or a key combo (Ctrl+Space).

Guard the dock API — in frontend/main/index.ts, wrap app.dock.* calls in if (process.platform === 'darwin')
(some are already guarded, check all of them).

udaykakade25:
Closes #24 

Adds a Windows-compatible path for global hotkey handling and platform-aware UI/STT behavior, while leaving the existing macOS hold-to-record flow untouched.

- main: on Windows, fall back to Electron's globalShortcut in toggle mode (press to start, press again to stop) since WinKeyServer.exe is not shipped on Windows; cancel hotkey recording on Windows and re-register the current accel since key-up detection is unavailable.
- main: make settings window titleBarStyle/trafficLightPosition macOS-only; default chrome on Windows.
- vite: expose process.platform to the renderer as a build-time constant so the renderer can branch on platform.
- onboarding: show platform-aware hotkey instructions (press-to-toggle on Windows, hold-to-record on macOS).
- streaming-stt: disable realtime streaming on Windows since the toggle hotkey flow doesn't pair with the Realtime API the same way the macOS push-to-talk flow does.

## Demo 

https://github.com/user-attachments/assets/74e06960-b87c-467d-92dc-50845e982215

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -175,6 +175,7 @@ function getAppWindowPosition(): { x: number; y: number } {
 function createAppWindow(): void {
   const { x, y } = getAppWindowPosition();
 
+  winToggleActive = false;
   mainWindow = new BrowserWindow({
     width: APP_WIDTH,
     height: APP_HEIGHT,
@@ -715,14 +716,11 @@ app.whenReady().then(() => {
       }
       keyListener = null;
     }
-    if (process.platform === "win32") {
-      globalShortcut.unregisterAll();
-      winToggleActive = false;
-    }
-
     // GlobalKeyboardListener requires WinKeyServer.exe which isn't shipped on Windows.
     // Hotkey recording is not supported on Windows — cancel and re-register.
     if (process.platform === "win32") {
+      globalShortcut.unregisterAll();
+      winToggleActive = false;
       settingsWindow?.webContents.send("hotkey-record:cancel");
       registerHotkey(currentHotkeyAccel ?? undefined);
       return;
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -8,6 +8,7 @@ import { serve } from "@hono/node-server";
 import {
   app,
   BrowserWindow,
+  globalShortcut,
   ipcMain,
   Menu,
   nativeImage,
@@ -41,6 +42,7 @@ let settingsWindow: BrowserWindow | null = null;
 let tray: Tray | null = null;
 let keyListener: GlobalKeyboardListener | null = null;
 let hotkeyPressed = false;
+let winToggleActive = false;
 let currentHotkeyAccel: string | null = null;
 
 /**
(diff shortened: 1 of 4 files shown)


apps/electron/src/main/index.ts
```diff
@@ -712,6 +715,18 @@ app.whenReady().then(() => {
...
+    // GlobalKeyboardListener requires WinKeyServer.exe which isn't shipped on Windows.
+    // Hotkey recording is not supported on Windows — cancel and re-register.
+    if (process.platform === "win32") {
+      settingsWindow?.webContents.send("hotkey-record:cancel");
+      registerHotkey(currentHotkeyAccel ?? undefined);
+      return;
+    }
 
     // Ensure binary is executable / signed before starting recording listener
     ensureMacKeyServerExecutable();
```
MathurAditya724, line 730: These two blocks could just be one — looks like they got split by accident during development. Something like:

```ts
if (process.platform === "win32") {
  globalShortcut.unregisterAll();
  winToggleActive = false;
  settingsWindow?.webContents.send("hotkey-record:cancel");
  registerHotkey(currentHotkeyAccel ?? undefined);
  return;
}
```

Not a big deal, just reads a bit odd having two consecutive platform checks right next to each other.

apps/electron/src/main/index.ts
```diff
@@ -1106,13 +1121,41 @@ function registerHotkey(hotkey?: string): void {
...
+        mainWindow?.webContents.send("hotkey:down");
+      } else {
+        winToggleActive = false;
+        mainWindow?.webContents.send("hotkey:up");
+      }
+    });
+    if (!registered) {
+      const errorPayload = {
+        message: `Could not register hotkey "${accel}". Try a different key combination in Settings.`,
+      };
```
MathurAditya724, line 1150: Minor edge case — if the main window gets closed and recreated while `winToggleActive` is `true`, the next hotkey press would send `hotkey:up` to a fresh window that never got the original `hotkey:down`. The user would have to press the hotkey twice to get back in sync.

Might be worth resetting `winToggleActive = false` when the main window is recreated, or when recording finishes through other paths (like a stop button in the UI). Not urgent but could save a confusing bug report down the line.

udaykakade25:
HI @MathurAditya724,
I have made the changes. 
Thanks!

PR #34, opened 2026-05-26
fix: reduce main-process event-loop pressure to prevent typing lag on macOS

MathurAditya724:
## Summary

Freestyle's `node-global-key-listener` intercepts **every keystroke system-wide** via a native binary (MacKeyServer) that blocks on the Node.js event loop for each event response. When the main process event loop is busy, keystroke responses are delayed, causing perceptible typing lag across all applications.

This PR reduces event-loop pressure through four targeted fixes:

### 1. Replace `execSync` with async `execFile` for frontmost-app detection (P0)
`getMacFrontmostApp()`, `getWindowsFrontmostApp()`, and `getLinuxFrontmostApp()` previously used `execSync` which **blocked the entire main thread** for up to 2-3 seconds on every hotkey press. During this time, the MacKeyServer binary would be waiting for keystroke responses that couldn't be processed, freezing all keyboard input system-wide.

Now uses an async `execFile` wrapper so the event loop remains free to process key events while the child process runs.

### 2. Throttle audio-level IPC from 60 fps to ~10 fps (P1)
During recording, `sendAudioLevel()` fired an IPC message on every `requestAnimationFrame` callback (~60/sec). The main process received each one and forwarded it to the settings window — 60 IPC serialization/deserialization cycles per second competing with the key listener's event responses.

Now throttled to ~10 fps (every 100ms), reducing IPC pressure by 6x with no visible difference to audio visualization.

### 3. Replace 60 fps RAF timer with 100ms setInterval (P1)
The recording timer called `setElapsed()` via `requestAnimationFrame` at ~60 fps, triggering a full React re-render each frame. A timer display only needs ~10 updates/sec.

Replaced with `setInterval` at 100ms, also fixed cleanup to use `clearInterval` instead of `cancelAnimationFrame`.

### 4. Ring buffer for AudioWorklet PCM processor (P1)
The `process()` callback runs ~344 times/sec and previously allocated a **new `Float32Array`** on every call, copying the old buffer + new data into it. This created heavy GC pressure on the audio thread.

Replaced with a pre-allocated ring buffer (~200ms capacity) that writes samples in-place with zero allocations during steady-state operation.

## Not included (future work)
- Replacing `node-global-key-listener` with an in-process native addon (eliminates the stdio round-trip entirely)
- Moving the Hono server to a UtilityProcess (isolates DB/network from the key listener event loop)

## How to verify
1. Launch Freestyle on macOS
2. Type in any application — keystrokes should feel responsive with no perceptible lag
3. Hold the hotkey to record — visualization, timer, and streaming should work normally
4. Release and verify transcription completes and pastes correctly

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -324,16 +324,39 @@ function showPill(): void {
   }
 }
 
-// -- macOS: Get frontmost app + browser tab context via AppleScript --
-function getMacFrontmostApp(): string | null {
-  try {
-    const { execSync } = require("node:child_process");
-    const appName = execSync(
-      `osascript -e 'tell application "System Events" to get name of first application process whose frontmost is true'`,
-      { encoding: "utf-8", timeout: 2000 },
-    ).trim();
+// -- Async helper: run a command without blocking the main thread --
+function execAsync(
+  cmd: string,
+  args: string[],
+  timeoutMs: number,
+): Promise<string> {
+  const { execFile } =
+    require("node:child_process") as typeof import("node:child_process");
+  return new Promise((resolve, reject) => {
+    execFile(
+      cmd,
+      args,
+      { encoding: "utf-8", timeout: timeoutMs },
+      (err, stdout) => {
+        if (err) reject(err);
+        else resolve((stdout as string).trim());
+      },
+    );
+  });
+}
 
-    // For browsers, try to get the active tab URL and title
+// -- macOS: Get frontmost app + browser tab context via AppleScript --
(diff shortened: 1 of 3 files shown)


MathurAditya724:
All CI checks passed (Lint, Typecheck, Build for Linux/macOS/Windows, Preview Changelog). Self-review found one minor style issue (inline require instead of top-level import) which was fixed in the follow-up commit. No logic bugs or test gaps identified — the ring buffer math, IPC throttling, and async conversion are all clean. Ready for human review.

PR #37, opened 2026-05-26
fix: idle-timeout mic release and first-run pill disappearing

MathurAditya724:
## Summary

Two fixes: restores fast recording start and fixes the first-run pill flash.

### 1. Idle-timeout mic release (fixes recording start lag)
After PR #35, every recording session called `releaseStream()` immediately, forcing a full `getUserMedia()` call (~100-300ms) on the next hotkey press. This undid the speed gains.

Now uses `scheduleRelease(30000)` instead — the mic stream stays alive for 30 seconds after recording ends. If the user presses the hotkey again within 30s, `hasLiveStream()` returns true and `acquireStream()` skips `getUserMedia()` entirely (0ms). After 30s idle, the stream is released and the OS mic indicator turns off.

- `Recorder.scheduleRelease(delayMs)` — starts a timer that calls `releaseStream()` after the delay
- `Recorder.cancelScheduledRelease()` — clears the timer (called by `acquireStream()`)
- `acquireStream()` cancels any pending release before checking the stream
- Abort paths (user cancelled during init) still use immediate `releaseStream()`

### 2. First-run pill disappearing
On the very first hotkey press, the Streamer singleton is created and opens a WebSocket to the server. If the server has no voice model configured or no API key, it sends an `error` message and closes the connection. The `onError` callback (after PR #36 fixes) would reset `wantsMicRef = false`, call `stopVisualization()`, and `hidePill()` — tearing down the recording that was just starting.

Fixed by guarding `onError`: only tear down recording state when `wantsMicRef.current` is true (i.e., during an active recording). Connection-time errors are silently ignored — the recording flow continues using the REST fallback path.

## Verification
- `pnpm run build` — passes (typecheck node + web, electron-vite build)
- `pnpm biome check` — passes
- Test: first hotkey press should show pill and enter recording (no flash/disappear)
- Test: repeated recordings within 30s should start instantly (no "initializing" state)
- Test: wait >30s after a recording, then press hotkey — should show "initializing" briefly, then record
- Test: check macOS menu bar — mic indicator should turn off ~30s after last recording

### DIFF
--- a/apps/electron/src/renderer/src/lib/recorder.ts
+++ b/apps/electron/src/renderer/src/lib/recorder.ts
@@ -2,11 +2,14 @@ import { encodeWavFromFloat32 } from "./wav";
 
 const TARGET_RATE = 16000;
 
+const RELEASE_DELAY_MS = 30_000;
+
 export class Recorder {
   private stream: MediaStream | null = null;
   private mediaRecorder: MediaRecorder | null = null;
   private chunks: Blob[] = [];
   private mimeType = "";
+  private releaseTimer: ReturnType<typeof setTimeout> | null = null;
 
   private hasLiveStream(): boolean {
     return (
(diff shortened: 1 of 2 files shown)


MathurAditya724:
Skipping — moving directly to native addon implementation (audify + uiohook-napi)

PR #57, opened 2026-05-27
fix: implement proper Sentry instrumentation for errors, traces, and telemetry

MathurAditya724:
Sentry was initialized but not instrumented — zero `captureException` calls existed in the entire codebase, no Hono error handler, no release tracking, no environment tags. This PR fixes the instrumentation so errors, traces, and telemetry actually flow to Sentry.

## Changes

**Electron main process:**
- Add `release` and `environment` to `Sentry.init()` for version correlation
- Set `platform`, `arch`, and `electron` version tags on app ready
- Add `Sentry.captureException` to server startup failures and key listener errors
- Flush pending events on quit via `Sentry.close(2000)` in `before-quit`

**Server (Hono):**
- Add global `app.onError()` handler that captures all unhandled route errors to Sentry
- Add `release` tag to standalone server `Sentry.init()`
- Export `captureException` helper from `sentry.ts` for use across routes
- Instrument transcription failures, post-processing LLM errors, and streaming session errors

**What was NOT added (intentionally):**
- No `@sentry/vite-plugin` for source maps — requires auth token configuration in CI secrets
- No session replay or profiling — heavy for a desktop voice tool
- No BrowserTracing in renderers — the pill must stay lightweight
- No blanket instrumentation of every `catch {}` — only critical error paths that indicate real failures

## Testing
Server typecheck and build pass. Main process typecheck passes. Pre-existing type error in `history.tsx` is unrelated.

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -1,8 +1,9 @@
 import * as Sentry from "@sentry/electron/main";
+import { app as _app } from "electron";
 
 Sentry.init({
   dsn: "https://b7ed8a9e5051cfe650f0f26ca2482b4b@o4509750817325057.ingest.us.sentry.io/4511454571528192",
-  release: `freestyle@${process.env.npm_package_version ?? "0.0.0"}`,
+  release: `freestyle@${_app.getVersion()}`,
   environment: process.env.NODE_ENV ?? "development",
   enabled: process.env.NODE_ENV === "production",
   skipOpenTelemetrySetup: true,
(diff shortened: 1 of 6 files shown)


MathurAditya724:
All CI checks green. Self-review caught and fixed 3 bugs: (1) Sentry.close() promise wasn't awaited on quit — events would be lost; (2) release tag used npm_package_version which is undefined in packaged Electron — switched to app.getVersion(); (3) server Sentry was active in development mode — added enabled guard matching Electron's behavior.

PR #75, opened 2026-05-28
feat: add local whisper.cpp voice model support

MathurAditya724:
## Summary
- Adds privacy-first local speech-to-text via bundled whisper.cpp binaries (macOS, Windows, Linux)
- Server-owned model download lifecycle with progress tracking, cancel, and retry — UI just polls `/api/whisper/status`
- New `local-whisper` transcription provider implementing the existing `TranscriptionProvider` interface with both batch (CLI) and streaming (whisper-server HTTP) paths

## Testing
- Server package type-checks cleanly (`tsc --noEmit`)
- Biome lint/format passes
- Pre-commit hooks pass
- No new type errors introduced in renderer (all errors are pre-existing Hono RPC client resolution)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ apps/electron/tsconfig.node.tsbuildinfo
 
 # Compiled native binaries (built locally and in CI)
 apps/electron/resources/bin/
+
+# Downloaded whisper.cpp binaries
+apps/electron/resources/whisper/
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+public-hoist-pattern[]=react
+public-hoist-pattern[]=react-dom
--- a/apps/electron/electron-builder.yml
+++ b/apps/electron/electron-builder.yml
@@ -18,6 +18,11 @@ extraResources:
       - "**/*"
 win:
   executableName: Freestyle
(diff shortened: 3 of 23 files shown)


tigerisaac:
Is it possible to add support for GPU platform acceleration for the model? MacOS has mlx, Nvidia has cuda, AMD has rocm.

It would be a lot more complex but from my experience with MLX, performance is significantly better, which is needed for an app like this.

MathurAditya724:
Yeah @tigerisaac that's step 2, cause those are complex I was thinking of adding them one by one

MathurAditya724:
Self-review complete. Lint and typecheck CI checks pass. Build jobs pending (platform packaging).

Changes reviewed and cleaned up:
- Removed all 34 debug log statements across 4 files
- Removed dead code (unused binary release definitions in constants.ts)
- Extracted shared `WHISPER_CPP_VERSION` constant and `webBodyToReadable()` helper to eliminate DRY violations
- Fixed .so file filter regex for versioned shared libraries
- Fixed download phase to correctly reflect whether binary build is needed
- Aligned GitHub org URL (`ggml-org`) across all files
- Moved dynamic imports to top-level static imports

Issue #66, opened 2026-05-28
ASR vocabulary biasing in dictionary


tigerisaac:
Currently, we run the dictionary as a regex parser. It works, but it should really be ASR vocabulary biasing.
Cloud providers support this (and in the future, local models could too). 

This would bridge the gap to using things like foreign names, terms, places, medical terms, etc.

We could also meet in the middle if dictionaries get too large, which would in theory reduce model performance.

matteo8p:
@tigerisaac Do you know how to do ASR vocabulary biasing? I agree, I think regex parsing is a naive approach. 

I'm wondering how we would handle that. Would love to have your help here.

tigerisaac:
Using the cloud providers, it would be pretty simple. Just passing a prompt through with the dictionary (i.e. Groq supports an initial_prompt parameter). I believe it's similar for local models as well if we plan to do that.

Something like "Key terminology to prioritize in this transcription: {item1}, {item2}, {item3}."
The prompt looks like an LLM prompt because many of the newer models use a speech LLM framework.
I don't know the exact list, but if older models don't support prompting like this then we use regex as a fallback, but I believe all mainstream ASR models will have at least a hotword list rather than a prompt.

I don't know if regex parsing needs to be removed with this change, I'll see if it's still helpful.

I'd be down to work on the feature

matteo8p:
@tigerisaac Is this issue safe to close?

tigerisaac:
Oh yeah it is, sorry about that I didn't know that I had to say "Closes issue" in the description lol

PR #82, opened 2026-05-29
ASR Vocabulary biasing and Streaming fixes

tigerisaac:
Added Vocabulary, a user-managed list of names and phrases that improves first-pass speech recognition (ASR), which is separate from Dictionary (our shortcuts after transcription). 

Features

- Vocabulary settings page (⌘4 in sidebar): add, edit, search (⌘K), import/export JSON, pagination
- Server API (`/api/vocabulary`): CRUD + bulk import/export backed by a `vocabulary` table
- Provider-specific ASR bias applied only on the transcription step (not post-processing):
  - OpenAI / Groq: context prompt with term list (char-budget capped)
  - Deepgram Nova 3: `keyterm` boosting (streaming uses a shorter list for WS limits)
  - Deepgram Nova 2: per-word `keywords` (multi-word phrases split)
  - ElevenLabs Scribe v2 realtime: `keyterms` on the WebSocket
  - ElevenLabs Scribe v1: no realtime bias — uses batch `/api/transcribe` instead

Streaming fixes

- Deepgram: reset session state per recording (not full reconnect); dedupe cumulative is_final segments so past dictations are not re-appended (bug I probably made but it's fixed lol)
- ElevenLabs: honor user commit vs auto-commit; use scribe_v2 realtime for WS; Scribe v1 falls back to batch when streaming is unsupported (if we tried streaming it times out)

How it differs from Dictionary

Vocabulary is ASR bias, so if you say something like "Maneesh", without vocabulary most models return "Manish". ASR biasing fixes that so that models will instead return "Maneesh". Dictionary is more like macros - "my address" becomes "1234 lane"

Things to test:
I don't have an OpenAI key, so that's still untested, but all other providers are working with the feature


tigerisaac:
This relates to #66

matteo8p:
@tigerisaac first, I REALLY appreciate you putting this feature up, this is so powerful and having ASR bias is a _must have_. I understand the difference between "vocabulary" and "dictionary" in your terms: 

- Vocabulary: mostly corrects spellings / terms (Victor -> Viktor) 
- Dictionary: entire keyword changes (my address -> 123 Beverly st)

Though functionally they're different, could we combine the two functionality under "Dictionary"? Maybe just extend Dictionary to include ASR bias? 

My argument for doing this is that the two features are close enough to combine into one imo. By combining into one, it's less strain on the user to understand the subtle difference. I think Wispr Flow does it really well, where they combine it under "Dictionary", and I think we should follow their approach there. 

https://github.com/user-attachments/assets/52aec2dc-1ad6-49e7-90a7-96445fbe49cf

tigerisaac:
Thanks for your response! From what I've seen with Freestyle, i think our dictionary looks and functions a lot more like Snippets.
I think the misspelling correction would be for homophones where I could change "Issac" to "Isaac", instead of just putting "Isaac" into the vocabulary, which is a thing I can add onto the vocabulary. Curious what you think of this.

tigerisaac:
We can definitely change names around but from what I'm understanding, we can aim to separate into two things - one to correct misspellings, add names and foreign words, and one to basically have shortcuts for. 
But if you're thinking of a different direction with how Freestyle handles it, I'm down to make changes

matteo8p:
> We can definitely change names around but from what I'm understanding, we can aim to separate into two things - one to correct misspellings, add names and foreign words, and one to basically have shortcuts for. But if you're thinking of a different direction with how Freestyle handles it, I'm down to make changes

Ahh you are right, Wispr does have Dictionary and Snippets as two separate things. I find it a bit silly haha because I feel like they're functionally the same. I can see why they branded it as two separate things, to guide users to use them differently. 

I feel a bit more indifferent now and really don't have a strong opinion. I think personally, the mental model of all of it being under dictionary makes sense to me, more simple. But that's just me. What is the mental model for you that makes sense? I want to choose the best for first time users to understand how to use Freestyle. 

Also, could they be the same on the backend? Could AST power _everything_ in Dictionary? Wondering what are its limitations. 

Let me know what you think! You have a better grasp of this feature than I do.

matteo8p:
From Discord: 

> Yep, my thoughts are:
> We could turn vocabulary into dictionary, and from my experience in 90% of cases I use them for different things
> For example I'll use dictionary to expand "my email" into an email address, "my house" to my house address and so on
> But I'll use vocabulary to get the names of my friends right
>

apps/electron/src/renderer/src/hooks/use-search-shortcut.ts
```diff
@@ -0,0 +1,23 @@
+import { type RefObject, useEffect } from "react";
+
+/** Focus the page search field on Cmd/Ctrl+K. */
+export function useSearchShortcut(
```
matteo8p: Nit, but I think we should just move this to vocabulary.tsx where it's being used. Don't think we should add hooks for single use

tigerisaac:
Oh yeah and to answer this: "Also, could they be the same on the backend? Could AST power everything in Dictionary? Wondering what are its limitations."

No, because the ASR prompt tells the model what words to expect. It's all sound data to them, but if I'm trying to say "Ibuprofen", the model without the prompt will hear "Eye bee pro fin" and reconstruct that with some other words. With ASR biasing it correctly identifies it. 
ASR biasing won't be able to tell the model to recognize "my email" and change that into "asdf@gmail.com". 

The way people use Snippets in WFlow (which is pretty much our dictionary) is to do huge blocks of text. This could be a prompt to AI, a link to their calendar, contact info, large formatted tables, and more.

apps/electron/src/renderer/src/hooks/use-search-shortcut.ts
tigerisaac: Got it, let me change that

matteo8p:
@tigerisaac Thanks for teaching me all this! I have a better understanding of it now.

tigerisaac:
No problem! Happy to help, I have a decent bit of experience on this since I've been fooling around with these models for a year or so lol

matteo8p (approved):
Ship

MathurAditya724 (commented):
## Review — ASR Vocabulary Biasing and Streaming Fixes

~1600 lines across 24 files. The vocabulary feature and streaming provider fixes are well-structured overall. The provider-specific bias dispatch in `vocabulary-bias.ts` is clean, and the ElevenLabs `commitTimeout` pattern is a good safety net. A few real bugs and some consistency gaps worth addressing.

---

### Bugs

**1. ElevenLabs bias path sends `_realtime` model ID to the batch API** — `elevenlabs.ts:93-94`, `transcribe-bias.ts:87-94`

When bias is present, `transcribe()` takes the early return on line 93-94, bypassing the `_realtime` stripping on lines 96-97. `transcribeElevenLabsWithBias` then calls `stripProviderPrefix(opts.model)` which yields `scribe_v2_realtime`, and passes that as `model_id` to the ElevenLabs batch `/v1/speech-to-text` endpoint. That endpoint does not accept `_realtime` suffixed model IDs. Strip `_realtime` before entering the bias path.

**2. Deepgram `commit()` has no timeout — can hang indefinitely** — `deepgram.ts:171-177`

ElevenLabs has `USER_COMMIT_TIMEOUT_MS` (line 298) so if the server never responds to the commit, `deliverUserFinal()` fires after the timeout. Deepgram's `commit()` sends `Finalize` and sets `commitRequested = true`, but if the Finalize response never arrives, `deliverFinal()` is never called and the user's commit hangs forever. Add a similar timeout.

**3. `/import` has no runtime validation, no size limit, no transaction** — `vocabulary.ts:149-179`

`c.req.json<...>()` is just a TypeScript type assertion. The body could be anything — an object, a string, an array of millions of items. There's no `zValidator`, no array length cap, and the insert loop isn't wrapped in a transaction. This means:
- Unbounded payload can block the event loop with synchronous SQLite inserts
- A crash mid-import leaves partial data with no rollback
- Performance is bad: each auto-committed insert is ~10x slower than batched in a transaction

Wrap in `BEGIN`/`COMMIT`, add a zod schema with `z.array(...).max(N)`, or at minimum a length check.

**4. Schema migration ordering: version 6 before version 5** — `schema.ts:92-116`

The `if (currentVersion < 6)` block (vocabulary) runs before `if (currentVersion < 5)` (format_rules). Works by coincidence (`CREATE TABLE IF NOT EXISTS`), but the reverse ordering is a maintenance trap. Swap them.

**5. `loadVocabularyTerms` silently swallows all errors** — `vocabulary.ts:14-23`

Any DB error returns `[]` with no logging. ASR bias silently stops working and nobody knows. At minimum log the error.

---

### Streaming

**6. Old session's `onPartial`/`onFinal` callbacks not guarded after reconnect** — `stream.ts:109-111`

`connectUpstream` closes over `onPartial`/`onFinal` callbacks. When a new session is created (reconnect), the old session's callbacks can still fire. The `onClose` callback has a guard (`if (upstream !== session) return` on line 202), but `onPartial` and `onFinal` do not. A late-arriving partial from the old Deepgram/ElevenLabs session would be forwarded to the client as if it came from the new session.

**7. OpenAI `commit()` silently drops when WebSocket is not open** — `openai.ts:116-118`

If the WebSocket disconnected, `commit()` returns silently and `onFinal` is never called. Deepgram and ElevenLabs both handle this by delivering a final with whatever text is accumulated. OpenAI should do the same — otherwise the client hangs waiting for a final that never arrives.

---

### Frontend

**8. Vocabulary import silently swallows all errors** — `vocabulary.tsx:183-198`

`catch { // ignore }` — if `JSON.parse` fails or the API rejects the import, the user gets zero feedback. Show a toast or set an error state.

**9. `deleteEntry` has no error handling** — `vocabulary.tsx:134-142`

If the `$delete` call throws, the promise rejection is unhandled. Wrap in try/catch or add `.catch()`.

**10. Deleting last entry on a page leaves user on an empty page** — `vocabulary.tsx:134-142`

After delete, `loadData()` re-fetches with the same `page` offset. If the deleted item was the only one on that page, the result is an empty list — even though earlier pages have entries. Adjust `page` when entries would be empty: `if (entries.length === 1 && page > 1) setPage(page - 1)`.

---

### Style / Consistency

**11. `DEEPGRAM_KEYTERM_MAX` used for ElevenLabs batch limit** — `vocabulary-bias.ts:155-157`

The non-streaming ElevenLabs path uses `DEEPGRAM_KEYTERM_MAX` (100). This is a Deepgram constant being repurposed. If it changes, ElevenLabs silently inherits the wrong value. Define a separate constant.

**12. `VocabularyRow` interface duplicated** — `routes/vocabulary.ts:9-15` vs. `lib/vocabulary.ts:3-9`

Identical interface in two files. Import from the lib to avoid drift.

**13. Hardcoded `⌘ K` on all platforms** — `vocabulary.tsx:249`

The keydown handler checks both `metaKey` and `ctrlKey`, but the visual hint always shows `⌘ K`. Show `Ctrl+K` on non-Mac. (Same issue exists in other pages — consistent but still wrong.)

**14. `isNova3Model` redundant condition** — `vocabulary-bias.ts:98-100`

`model.includes("nova-3-general")` is always true when `model.includes("nova-3")` is true. The second check is dead code.

Issue #84, opened 2026-05-29
Re-binding hotkey is broken


tigerisaac:
Found a bug that doesn't let you rebind hotkey in settings.
I believe the cause is that the recording IPC implementation was removed in #50 and a replacement was never added.

Submitting a PR to address this and some other features soon (toggle vs hold mode for recording?).

PR #85, opened 2026-05-29
Fix/hotkey rebinding

tigerisaac:
Hotkey rebinding is now fixed. Issue was introduced with #50
#50 replaced node-global-key-listener with native binaries for push to talk, but removed recording listeners which were used in key rebinding. 

This PR restores rebinding by spawning a temporary listener while rebinding, made for all platforms.
Additionally, it adds a feature to be able to switch between holding and toggling for recording mode. Hold mode makes you hold down the key until you're done talking, toggle allows you to turn it on and keep it on until you press the hotkey again to turn it off.

Addresses #84

### DIFF
--- a/apps/electron/native/linux-key-listener.c
+++ b/apps/electron/native/linux-key-listener.c
@@ -39,6 +39,7 @@ static int g_requireAlt = 0;
 static int g_requireShift = 0;
 static int g_requireMeta = 0;
 static int g_modifiersOnly = 0;
+static int g_record_mode = 0;
 
 /* Tracked modifier state */
 static int g_ctrlDown = 0;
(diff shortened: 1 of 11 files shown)


MathurAditya724:
Thanks for this man! And special thanks for "holding and toggling for recording mode", was thinking about adding that.

can you resolve the merge conflicts? Then I will take a look

tigerisaac:
Yeah I’ll get on it in a bit I’ll let you know when I do that

tigerisaac:
Alright fixed them

MathurAditya724 (changes requested):
Nice work restoring hotkey rebinding and adding hold/toggle mode. A few issues to address before this can be merged:

## Bugs

### 1. Settings window closed during recording kills global hotkey permanently
`hotkey-record:start` (index.ts:737-740) stops the key listener and nulls it. If the user closes the settings window mid-recording (Cmd+W), neither `hotkey-record:stop` nor `onCancel` fire — the key listener is never re-registered. The global hotkey is dead until app restart.

**Fix**: In the `settingsWindow` `closed` handler (around line 247), check if the recorder is active and call `stopHotkeyRecorderProcess()` + `registerHotkey()`.

### 2. Recorder binary crash leaves UI stuck in "recording" state
In `hotkey-recorder.ts:101`, the `close` handler only nulls the process. If the native binary crashes during recording, no cancel/error event reaches the renderer. The settings UI stays on "Press a key combination..." permanently.

**Fix**: In the `close` handler, if `this.target` is still set (meaning `stop()` wasn't called intentionally), call `this.sendCancel()`.

### 3. Hold mode broken on globalShortcut fallback
The `globalShortcut.register` fallback (index.ts:1121) only calls `handleNativeHotkeyDown()`. In hold mode, the down event starts recording but there's no key-up detection — recording never stops.

**Fix**: Force toggle mode when using the globalShortcut fallback, or log a warning.

### 4. `updateHotkey()` race in key-listener.ts
`updateHotkey()` (line 387) sets `destroyed=false` then kills the process. If the killed process exits non-zero, the `close` handler fires `scheduleRestart()` concurrently with the explicit `start()` at line 407 — two listener processes can run simultaneously.

**Fix**: Remove the old process's event listeners before killing, or set a flag to skip `scheduleRestart` during intentional restarts.

### 5. Unnecessary hotkey re-registration on settings page navigation
`use-hotkey-recorder.ts:255` calls `stopHotkeyRecording()` unconditionally on unmount, even if recording was never started. This triggers `registerHotkey()` → native binary restart every time the user navigates to/from the settings page.

**Fix**: Track whether recording was started via a ref, only call `stopHotkeyRecording()` if it was.

## Scope

The `before-quit` handler at index.ts:1163 adds a fetch to `/api/whisper/server/stop` which is unrelated to hotkey rebinding. This was merged into main via #75 — either remove this line (it's already on main) or rebase to avoid the duplicate.

## Minor

Windows native code (windows-key-listener.c:425-436) has inconsistent indentation in the modifiers-only block.

MathurAditya724:
## Inline Review Comments

Since GitHub's API is being difficult with inline diff comments, here are the findings with exact file locations:

---

### 1. Bug: Settings window close during recording kills hotkey permanently
**`apps/electron/src/main/index.ts:737-740`**

`hotkey-record:start` stops the key listener and nulls it. If the user closes the settings window mid-recording (Cmd+W), neither `hotkey-record:stop` nor `onCancel` fire — the key listener is never re-registered. The global hotkey is dead until app restart.

**Fix**: In the `settingsWindow` `closed` handler (around line 247), check if the recorder is active and call `stopHotkeyRecorderProcess()` + `registerHotkey(currentHotkeyAccel ?? undefined)`.

---

### 2. Bug: Recorder binary crash leaves UI stuck in recording state
**`apps/electron/src/main/hotkey-recorder.ts:101-103`**

The `close` handler only nulls the process:
```ts
proc.on("close", () => {
  this.process = null;
});
```
If the native binary crashes during recording, no cancel/error event reaches the renderer. The settings UI stays on "Press a key combination..." permanently.

**Fix**: Check if `this.target` is still set (meaning `stop()` wasn't called intentionally). If so, call `this.sendCancel()` to notify the renderer.

---

### 3. Bug: Hold mode broken on globalShortcut fallback
**`apps/electron/src/main/index.ts:1117-1125`**

The `globalShortcut.register` fallback only calls `handleNativeHotkeyDown()`:
```ts
const registered = globalShortcut.register(accel, () => {
  handleNativeHotkeyDown();
});
```
In hold mode, the down event starts recording but there's no key-up detection — recording starts and never stops.

**Fix**: Force toggle mode when using the globalShortcut fallback, or warn the user that hold mode requires the native binary.

---

### 4. Bug: Race condition — two listener processes can run simultaneously
**`apps/electron/src/main/key-listener.ts:387-407`**

`updateHotkey()` sets `destroyed=false` then kills the process. If the killed process exits non-zero, the `close` handler at line 155 fires `scheduleRestart()` concurrently with the explicit `start()` at line 407. Both spawn a new process.

**Fix**: Remove the old process's event listeners before killing, or set a flag to skip `scheduleRestart` during an intentional restart.

---

### 5. Bug: Unnecessary hotkey re-registration on every settings page navigation
**`apps/electron/src/renderer/src/hooks/use-hotkey-recorder.ts:255-258`**

The unmount cleanup calls `stopHotkeyRecording()` unconditionally, even if recording was never started. This triggers `registerHotkey()` → native binary restart every time the user navigates to/from the settings page.

**Fix**: Track whether recording was started via a ref. Only call `stopHotkeyRecording()` on unmount if recording was actually active.

---

### 6. Scope: Unrelated change
**`apps/electron/src/main/index.ts:1163`**

The whisper server stop in `before-quit` was already merged into main via #75. Either rebase onto current main to pick it up, or remove this line to keep the PR focused on hotkey rebinding.

tigerisaac:
Thanks for looking through it. I've updated everything accordingly

MathurAditya724 (approved):
All 5 bugs from the previous review are fixed correctly:

1. **Settings window close** — `settingsWindow.on('closed')` now cleans up the recorder and re-registers the hotkey.
2. **Recorder crash recovery** — `close`/`error` handlers check `this.target !== null` and call `sendCancel()` on unexpected exits. Verified `sendCancel()` is safe after `this.process = null`.
3. **globalShortcut fallback** — inline toggle logic replaces `handleNativeHotkeyDown()`, so hold mode can't leave recording stuck.
4. **updateHotkey race** — `proc.removeAllListeners()` before kill prevents `scheduleRestart` from racing with the explicit `start()`. Verified this only strips process-level events, not stdout/stderr data listeners (those are on separate stream objects).
5. **Unmount cleanup** — `recordingActiveRef` gates the `stopHotkeyRecording()` call, preventing unnecessary re-registration on settings navigation.

The whisper server stop in `before-quit` is a minor scope overlap with #75 but harmless (idempotent). No new issues found in the fix commit.

PR #86, opened 2026-05-29
fix: cleanup LLM model selection and provider handling

### LINKED ISSUE #78: Groq model list for LLM cleanup shows deprecated models and drops required provider prefix
The Groq model picker in Settings → Models → LLM cleanup has two issues that cause requests to fail silently:
1. Deprecated models are listed — Models that Groq has deprecated (e.g. gemma2-9b-it) still appear in the picker. Selecting them results in a failed cleanup request with a 400 error from Groq's API.
2. Missing provider prefix in model ID — Some Groq models require the full model ID including the provider namespace prefix (e.g. qwen/qwen3-32b). Currently only the short model name is sent, so the request fails silently and the error is only visible in logs.

### Steps to reproduce:
1. Open Settings → Models
2. Enable LLM cleanup
3. Select a Groq model (e.g. a deprecated one, or one that needs a prefix)
4. Record audio and trigger transcription with cleanup enabled
5. Observe that cleanup silently fails; check logs for the error

### Expected behavior:
- Deprecated models should be filtered out of the Groq model list
- Model IDs should include the required provider prefix when sent to the Groq API

### Actual behavior:
- Deprecated models appear selectable but fail on use
```
 error: {
      message: 'The model `gemma2-9b-it` has been decommissioned and is no longer supported. Please refer to https://console.groq.com/docs/deprecations for a recommendation on which model to use instead.',
      type: 'invalid_request_error'
    }
```

- Short model names (without prefix) are sent, causing 404 / "model not found" errors
```
 error: {
      message: 'The model `qwen3-32b` does not exist or you do not have access to it.',
      type: 'invalid_request_error'
    }
```

### Model Reference:
 [Groq model list](https://console.groq.com/docs/models)

xikhar:
This fixes #78 

  - Add Mistral chat provider support.
  - Filter deprecated and cleanup-unsuitable models from the LLM picker.
  - Preserve Groq namespaced model IDs while normalizing legacy provider-prefixed IDs for other providers.
  - Use provider-aware model cost lookup.
  - Strengthen cleanup prompt and strip Qwen `<think>` leakage.
  - Add timeout for `models.dev` registry fetches.
  
>Note: `https://models.dev/api.json` remains the source of truth and bottleneck for model/provider metadata. If that data is stale or incorrect, our filtering and availability can only be as accurate as the registry.

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -18,6 +18,7 @@
     "@ai-sdk/elevenlabs": "^2.0.33",
     "@ai-sdk/google": "^3.0.79",
     "@ai-sdk/groq": "^3.0.39",
+    "@ai-sdk/mistral": "^3.0.37",
     "@ai-sdk/openai": "^3.0.65",
     "@freestyle/validations": "workspace:*",
     "@hono/mcp": "^0.3.0",
--- a/apps/server/src/lib/post-process.ts
+++ b/apps/server/src/lib/post-process.ts
@@ -1,5 +1,5 @@
 import { generateText } from "ai";
-import { getModelCost } from "../routes/models.js";
+import { getModelCost, isCleanupModelSupported } from "../routes/models.js";
 import { getDb } from "./db.js";
 import { createChatModel, getDefaultModels } from "./providers.js";
 import { captureException, metrics } from "./sentry.js";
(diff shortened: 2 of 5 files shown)


matteo8p (approved):
Ship it

matteo8p:
@shikhargen Merged!!

PR #92, opened 2026-05-30
fix: fallback to REST transcription when OpenAI Realtime API is unsupported

udaykakade25:
## Summary

  Problem: When using gpt-4o-transcribe or gpt-4o-mini-transcribe models, pressing Alt+Space
  showed "Realtime Beta API not supported" and no text was pasted. This happened because these
  models route through OpenAI's Realtime API (/v1/realtime), which requires special beta access
  that not all API keys have.

  ---
  apps/server/src/routes/stream.ts

  - On Realtime API error, now sends {type: "config", streaming: false} to the client before the
  error message, telling it to switch to REST mode
  - Added streamingUnsupported flag — once the error fires, the server stops retrying the
  Realtime API connection (prevents the error from firing multiple times)
  - Guarded both reconnect sites (onClose and "start" message handler) with
  !streamingUnsupported

  apps/electron/src/renderer/src/pages/app.tsx

  - In the onError handler: if a recording was already committed and awaiting a result, grabs
  the accumulated PCM audio via streamer.getWavBlob() and POSTs it to the REST /api/transcribe
  endpoint — text gets pasted silently with no error shown
  - Added if (!useStreamingRef.current) return as a final guard to suppress any stray error
  messages once streaming has been disabled

  ---
  Result: First press now works — text is pasted silently. All subsequent presses use REST
  directly. Fix applies to all platforms (Windows, macOS, Linux).

### DIFF
--- a/apps/electron/native/windows-key-listener.c
+++ b/apps/electron/native/windows-key-listener.c
@@ -202,11 +202,15 @@ LRESULT CALLBACK LowLevelKeyboardProc(int nCode, WPARAM wParam, LPARAM lParam) {
                     printf("KEY_DOWN\n");
                     fflush(stdout);
                 }
+                /* Suppress the key event so Windows doesn't process it
+                   (e.g. Alt+Space would otherwise open the system window menu) */
+                if (g_isKeyDown) return 1;
             } else if (isUp) {
                 if (g_isKeyDown) {
                     g_isKeyDown = FALSE;
                     printf("KEY_UP\n");
                     fflush(stdout);
+                    return 1;
                 }
             }
         }
--- a/apps/electron/native/windows-key-listener.c
+++ b/apps/electron/native/windows-key-listener.c
@@ -318,8 +318,10 @@ LRESULT CALLBACK LowLevelKeyboardProc(int nCode, WPARAM wParam, LPARAM lParam) {
                     printf("KEY_DOWN\n");
                     fflush(stdout);
                 }
-                /* Suppress the key event so Windows doesn't process it
-                   (e.g. Alt+Space would otherwise open the system window menu) */
+                /* Suppress the key event globally so Windows doesn't process it.
+                   This fires on both the initial press and held-key repeats.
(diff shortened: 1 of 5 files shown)


MathurAditya724 (commented):
Good fix overall — the streaming fallback approach is sound and the Windows key suppression is correct.

A few things:

**app.tsx `onError` handler — `getWavBlob()` clears the PCM buffer**

`streamer.getWavBlob()` (streamer.ts:111-121) resets `pcmChunks` and `pcmSampleCount` to zero after building the blob. If `onError` fires but the REST fetch fails and the resolver was already consumed, the audio is gone. That's fine for the current flow (resolver is consumed once), but worth noting that `getWavBlob()` is destructive — it can only be called once.

**`if (g_isKeyDown) return 1` placement in KEY_DOWN path**

The `return 1` fires on every KEY_DOWN event after the initial one (repeat suppression), but it also fires on the **initial** KEY_DOWN event because `g_isKeyDown` was just set to TRUE two lines above. This means the hotkey itself is consumed and never reaches any other app. That's the intended behavior for Alt+Space (prevent system menu), but if the hotkey is changed to something that other apps also need, suppressing it globally could be surprising. Current behavior is correct for the default hotkey though.

**stream.ts — `streamingUnsupported` persists for the entire WebSocket session**

Once `streamingUnsupported` is set to `true`, the server never reconnects upstream for the lifetime of that WebSocket. The client-side WebSocket reconnects on close (streamer.ts:219-227), which creates a new server-side handler with a fresh `streamingUnsupported = false`. So the flag effectively resets when the WebSocket reconnects. This is correct.

**Minor: `onError` moved `pillActiveRef.current` check below resolver check**

The original code had `if (!pillActiveRef.current) return` as the first line. The new code checks the resolver first. This means if the pill is inactive but a resolver exists, the REST fallback still runs and resolves the promise. This is actually better — it prevents a dangling unresolved promise.

No blocking issues.

PR #96, opened 2026-05-30
feat: unify model picker cards

xikhar:
This updates the LLM Model Picker UI to be consistent with the Voice Models UI update.

Changes: 
  - Add shared LLM model row and provider header components.
  - Update Settings LLM picker to use shared card rows and provider-aware filters.
  - Update Onboarding text cleanup picker to match the voice model card style.
  - Widen onboarding model-selection steps for better selector spacing.
  - Fix onboarding selector scrollbars clipping outside rounded containers.
  
  
  <p align="center">
image
</p>
<p align="center">
image
</p>

### DIFF
--- a/apps/electron/src/renderer/src/pages/settings/models.tsx
+++ b/apps/electron/src/renderer/src/pages/settings/models.tsx
@@ -5,6 +5,11 @@ import {
   localLlmConfigSchema,
 } from "@freestyle/validations";
 import { zodResolver } from "@hookform/resolvers/zod";
+import {
+  LlmModelRow,
+  PROVIDER_FILTER_MARKS,
+  ProviderModelHeader,
+} from "@renderer/components/model-row";
 import { Toggle, VoiceRow } from "@renderer/components/voice-row";
 import { getApiBase, getClient } from "@renderer/lib/api";
 import {
@@ -19,7 +24,6 @@ import {
 import { cn } from "@renderer/lib/utils";
 import {
   AlertTriangle,
-  Check,
   ChevronRight,
   CircleDollarSign,
   Cloud,
@@ -29,6 +33,7 @@ import {
   Key,
   Laptop,
   Loader2,
+  type LucideIcon,
   Mic,
   Pencil,
   Plus,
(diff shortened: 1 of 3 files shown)


apps/electron/src/renderer/src/pages/settings/models.tsx
```diff
@@ -1284,70 +1372,67 @@ function LlmPicker({
...
-              ? models.filter(
-                  (m) =>
-                    m.model_name.toLowerCase().includes(q) ||
-                    m.model_id.toLowerCase().includes(q) ||
-                    providerName.toLowerCase().includes(q),
-                )
-              : models;
-            if (filtered.length === 0) return null;
+      {filter === "all"
+        ? visibleProviderGroups.map(({ providerId, providerName, models }) => (
```
MathurAditya724: The old code had `.slice(0, 20)` on each provider's filtered model list. That cap is gone now — all models render. Fine for typical provider catalogs, but could cause DOM bloat if a provider exposes hundreds of models. Was the removal intentional? If so, worth a note in the PR description.

apps/electron/src/renderer/src/pages/settings/models.tsx
```diff
@@ -1284,70 +1372,67 @@ function LlmPicker({
...
-                })}
-              </div>
+              <LlmModelRow
+                key={`${providerId}:${model.model_id}`}
+                name={model.model_name}
+                providerName={providerName}
+                modelId={model.model_id}
+                selected={isActive}
+                hasKey={keyProviders.has(providerId)}
+                first={!showLocal && index === 0}
```
MathurAditya724: `!showLocal` is dead code here — this branch only executes when `filter` is a specific cloud provider, which always sets `showLocal = false`. Could simplify to `first={index === 0}`.

MathurAditya724 (commented):
Clean refactor. The `ModelPickerShell` extraction consolidates ~80 lines of duplicated chrome into a reusable component with good slot-based composition (`banner`, `headerAccessory`, `empty`). The selected-model check in onboarding now verifying both `provider_id` and `model_id` is an improvement over the old code which only checked `model_id`. Scrollbar fix is correctly implemented. Two minor notes inline.

xikhar:
Updates: 

- Add paged LLM rows with `Show more` controls to avoid rendering every provider model at once.
- Removed dead code

image

image

apps/electron/src/renderer/src/pages/settings/models.tsx
xikhar: Now it can handle that gracefully by a simple pagination system that shows only 20 models initially, and the user can click Show More to load further.

MathurAditya724 (approved):
Both findings addressed — pagination with `MODEL_ROW_PAGE_SIZE` + "Show more" button is a nice upgrade over the old hard `.slice(0, 20)`, and the dead-code guard is cleaned up. LGTM.

PR #97, opened 2026-05-30
fix: update & restart button and download state tracking

MathurAditya724:
## Summary

- **Fix "Restart & Update" doing nothing**: The `before-quit` handler was calling `event.preventDefault()` + `app.exit(0)`, which short-circuited `autoUpdater.quitAndInstall()` — the electron-updater never got to launch the installer. Added an `isUpdaterQuitting` flag so the quit proceeds normally when triggered by the updater.
- **Fix settings page not reflecting background downloads**: When auto-download was enabled, the download would start in the background but the settings page still showed a "Download" button instead of "Downloading...". Now the main process tracks download state (`idle`/`downloading`/`downloaded`) and returns it from `updater:check`, plus sends a `updater:downloading` IPC event so the UI stays in sync.
- **Add error handling for update failures**: Added `autoUpdater.on("error", ...)` handler that forwards errors to the renderer via `updater:error` IPC, so the UI can recover from failed downloads (clears the "Downloading..." state, shows the error message) instead of getting stuck indefinitely.

## Changed files

| File | Changes |
|---|---|
| `apps/electron/src/main/index.ts` | Track `updateDownloadState`, add `isUpdaterQuitting` flag, add error handler, update `updater:check` return shape, fix `before-quit` |
| `apps/electron/src/preload/index.ts` | Update `checkForUpdate` return type, add `onUpdateDownloading` and `onUpdateError` bridges |
| `apps/electron/src/preload/index.d.ts` | Update type declarations to match |
| `apps/electron/src/renderer/src/pages/settings/general.tsx` | Handle new `checkForUpdate` return shape, subscribe to downloading/error events, show error in UI |


MathurAditya724:
Closing — branch was based off the wrong base. Will re-open from main.

Issue #83, opened 2026-05-29
Follow-up on #75 to add MLX platform support


tigerisaac:
Since per-platform architecture is too much to do in one issue, I figured we'd tackle the hardest one first.
I'd like to add Apple M-series only STT path which would have these features gated by hardware detection (basically all the features described will only affect MacOS on Silicon chips). We could check RAM automatically to get the right model tier (tiny, small, etc) and this issue would currently only do that for Whisper models. Later on we could probably get language detection in to find best models to recommend.

Motivation:
Better ASR performance on M series than current whisper.cpp in #75 

Features & scope:
Add detection for framework 
RAM based model tiers
MLX implementation of TranscriptionProvider

Some things to think about with local model support:
I didn't read fully into #75 so I'm unsure if this is already addressed, but we should figure out how we want models to be activated. I'm thinking of two paths:
- Cold start each time, hopefully the transcription speed to RT ratio (forgetting the acronym for this) will beat out the start time.
- Cold start but keep model warm and a X minute window on each activation until model is fully unloaded

1st path is simple and least RAM usage, but potential latency problems.
2nd path seems like a good compromise so that users don't constantly face cold starting on each transcription.

RAM to model defaulting would also just be like a strong suggestion, not a requirement.

MathurAditya724:
Yeah I would suggest the 2nd path, that's what I did for whisper.cpp

tigerisaac:
Got it, I'll experiment with this a little. Maybe branch out of whisper.cpp (i.e. English-only users would in theory see much better results from Parakeet).

matteo8p:
I'm going to close out this issue.

PR #112, opened 2026-06-01
MLX local models with Qwen3

tigerisaac:
Addressing #83, here is an implementation of Qwen3 ASR for MLX on M-series MacBooks!

Let's start with caveats:
MLX is a whole framework that runs on Python. Integration with it is not a small step, requires around 400mb of extra storage space - by bundling Python we don't require users to have Python installed already. Instead of bundling the worker that powers MLX models into the app for all users, I instead introduce a GH actions workflow that can be found in Actions -> MLX ASR Runtime. 

Someone with write access will need to publish this with publish=true on main so that the mlx-asr-worker-v1 exists.

Technical things:
A backend server runs: MLX ASR worker lifecycle (`mlx-asr` module), stdio Python worker in dev, optional frozen `mlx_asr_worker` in production, `local-mlx` streaming/batch provider, `/api/mlx-asr/*` routes, vocabulary bias passed as well.

Model sources are from HuggingFace. Streaming is employed to generate faster responses, about 1.5s partials on 16kHz PCM. We also employ a slider for users to be able to keep models warm between requests (on my machine this is barely a difference, but some would probably benefit).

We additionally employ platform gating so features are only seen and used by the correct platform. Also protects against Intel Macs mistakenly trying to use MLX.

Qwen supports the Vocabulary feature with word boosting, which is why the Vocabulary PR commits are included in this. A little messy but I wanted to release this.

Tested on local dev and also distribution build. Works on both, and I'm seeing some crazy startup times (unsure why they're so low but I'll take it). Hopefully can be validated on other machines.

Final thoughts: I think this PR will need a lot more thinking and testing, maybe if we start a beta branch this can be put on it. I went to the direction of not requiring a Python installation because some people just never touch code, and it would be much more seamless for them. That introduces the problem of the app becoming much larger, which can kind of be counteracted with our efforts (Also, if they have Python 3.12 with mlx-audio already, it won't download the worker). But it is a cool feature, it's kind of insane to reach this latency while not sacrificing RAM or storage space.


MathurAditya724 (commented):
## Review — MLX Local Models with Qwen3

Big feature, ~5k lines across 50 files. I focused on correctness, security, and process lifecycle. There's good work here — the generation-counter approach in `MlxLocalStreamingSession` is solid, and the stdin/stdout IPC model is a reasonable choice. That said, there are several issues worth addressing before merge.

---

### Bugs

**1. `stopMlxServer` fallback kill sends SIGTERM again, not SIGKILL** — `server.ts:503-505`

The 5-second fallback timeout calls `proc.kill()` (SIGTERM) — the same signal already sent on line 519. A stuck Python process (e.g., hung in GPU computation) won't respond to a second SIGTERM. This should be `proc.kill('SIGKILL')`.

**2. Race condition in `ensureMlxServerRunning` on concurrent model switches** — `server.ts:105-127`

If two callers invoke this with different model IDs concurrently, both pass the guard (different `currentModelId`), both `await stopMlxServer()`, and both call `startWorker`. The second overwrites `workerProcess`, orphaning the first child. Need a mutex or a check-after-await pattern.

**3. Schema migration ordering: version 6 block before version 5** — `schema.ts:92-116`

The `if (currentVersion < 6)` block (vocabulary table) appears before `if (currentVersion < 5)` (format_rules). Both use `CREATE TABLE IF NOT EXISTS` so it works today, but the reversed ordering is a maintenance hazard — anyone reading this code or adding a migration between versions 5 and 6 will be confused.

**4. `vocabulary-bias.ts` uses `DEEPGRAM_KEYTERM_MAX` for ElevenLabs batch** — `vocabulary-bias.ts:157`

Copy-paste: the non-streaming ElevenLabs path uses `DEEPGRAM_KEYTERM_MAX` (100) as its term limit. If the Deepgram constant changes, ElevenLabs silently inherits the wrong limit. Define `ELEVENLABS_BATCH_KEYTERM_MAX` separately.

**5. Vocabulary `/import` has no runtime validation or size limit** — `vocabulary.ts:149-179`

`c.req.json<...>()` is a TypeScript type assertion, not runtime validation. A client can POST millions of entries — each one a blocking synchronous SQLite insert in a loop with no transaction wrapper. This is both a DoS vector (blocking the event loop) and a data-integrity issue (partial import on crash). Wrap in a transaction and add a `zValidator` schema with a max-length array.

**6. `mlx_asr_server.py` `on_partial` lambda ternary is fragile** — `mlx_asr_server.py:205-211`

Python's ternary inside a lambda parses as a single lambda with a conditional body, not two separate lambdas selected by a ternary. It works by coincidence (the captured bool doesn't mutate), but is confusing and will break subtly if the code evolves. Add parentheses around each lambda branch.

---

### Process Lifecycle

**7. `unloadTimer` not `unref()`'d** — `server.ts:467`

The keep-alive timer (up to 10 minutes) will prevent the Node.js process from exiting. In Electron, closing the window won't terminate the process until the timer fires. Add `.unref()`.

**8. No `process.on('exit')` handler to stop the worker** — `server.ts` (module-level)

If the Node process exits (crash, SIGTERM, Electron quit), the Python worker can be orphaned. It will _usually_ self-terminate via stdin EOF, but not if it's mid-inference. Register a cleanup handler.

---

### Build / Packaging

**9. `sign_mac_app.sh` contradicts its own warning** — `sign_mac_app.sh:5-8, 30-31`

The header says "`codesign --deep` on the whole .app can break Electron Framework Team ID matching and prevent the app from launching." Then line 31 does exactly that: `codesign --deep --force --sign - "${APP}"`. The `--force` flag overwrites Electron's existing valid signatures. This can render the app unlaunchable on macOS Sequoia+.

**10. `mlx_asr_server.py` bundled as `extraResources` for all platforms** — `electron-builder.yml:19-20`

The MLX server Python script is in the top-level `extraResources`, not under `mac:`. It gets bundled into Windows and Linux builds where MLX cannot run. Move it under the `mac:` section (line 43).

**11. Unpinned pip dependencies** — `build_mlx_asr_worker.sh:19`

`pip install -U pyinstaller mlx-audio "huggingface_hub[hf_xet]"` with no version pins makes builds non-reproducible. A breaking change in any dependency silently breaks the CI workflow.

---

### Style / Consistency

**12. Vocabulary and MLX API calls use raw `fetch` instead of the typed Hono RPC client** — `models.tsx`, `onboarding.tsx`

The codebase has `getClient()` for typed API calls, but all Whisper/MLX endpoints use string-interpolated `fetch`. Bypasses type safety and is inconsistent with the rest of the API surface.

**13. Duplicate `VocabularyRow` interface** — `routes/vocabulary.ts:9-15` vs. `lib/vocabulary.ts:3-9`

Identical interface defined in two places. The route should import from the lib.

---

### Docs

**14. `mlx-asr-packaging.md` uses PR language** — line 94: "This PR keeps the DMG smaller..." Should say "This approach" in committed docs.

**15. `mlx-asr-packaging.md` resolution order is incomplete** — lists 6 steps but `python.ts` searches many more paths including a 12-level upward `__dirname` traversal. Doc should match code.

tigerisaac:
Addressed issues and redid the docs. My bad for missing the docs I don't really know how they got shoved in. Made a new guide for testing MLX locally to replace them

PR #117, opened 2026-06-01
Add basic telemetry

matteo8p:
# Overview

Hey team, we're planning to add PostHog telemetry. I understand that telemetry is sensitive and want to be transparent on how we're approaching this. 

Things we plan to record: 
Basic actions. Transcription complete / failures. Which models selected and used etc.
Random assigned anon ID for users. Helps us track # of users using Freestyle

Things we do not record: 
Any personally identifying information
Transcript contents, audio recordings. The actual vocabulary terms, etc. The contents you're putting into Freestyle

Telemetry is important to track bugs and failures, get a sense of usage, figure out what features ppl like and don't. It helps us with insights on improving the product. However, privacy is the whole point of the project, and we will never record any form of PII. 

# Exact events

- transcription completed
- streaming transcription completed
- transcription failed
- api key configured
- api key deleted
- model configured
- default model changed
- model deleted
- whisper model download started
- whisper model deleted
- vocabulary term added
- vocabulary term deleted
- vocabulary terms imported
- setting updated
- history cleared


MathurAditya724 (changes requested):
## Review

The instrumentation coverage is solid — events are placed at the right call sites and properties are relevant. A few things need to be addressed before merging.

### Must fix

**1. No opt-out mechanism — `apps/server/src/lib/posthog.ts`**

Telemetry fires unconditionally in all environments. Users have no way to disable it — no setting toggle, no env var, no `DO_NOT_TRACK` header support. For a project whose "privacy is the whole point," shipping always-on analytics without user consent or opt-out is a problem. At minimum:
- Add a `telemetry_enabled` setting (default `true` is fine) and a UI toggle.
- Check it in `capture()` / `captureException()` before sending.
- Respect the [Console Do Not Track](https://consoledonottrack.com/) convention (`DO_NOT_TRACK=1` env var).

**2. PostHog API key hardcoded in source — `apps/server/src/lib/posthog.ts:6`**

```ts
const POSTHOG_API_KEY = "phc_mDhFafyLK3Safsrrehi7rnH2X9jVMMGNAwKWuJsEN54w";
```

This is a write key baked into the client binary, so it's not a secret in the traditional sense — PostHog client keys are designed to be public. However, hardcoding it makes it impossible to rotate without a release, and it means dev/CI builds also send events to your production project. Consider injecting it at build time or at least gating it behind `getEnvironment() === "production"` so dev sessions don't pollute your data.

**3. `require()` inside an ESM module — `apps/server/src/lib/posthog.ts:31`**

```ts
const { getDb } = require("./db.js") as typeof import("./db.js");
```

Using `require()` in what appears to be an ESM codebase (all other files use `import`) is fragile. If the build config changes to pure ESM this will break at runtime. Use a dynamic `import()` or restructure so `getDeviceId()` receives the db as a parameter, which also makes it testable.

### Should fix

**4. `getClient()?.capture(...)` — optional chaining on a non-nullable return — `posthog.ts:60,75`**

`getClient()` always returns a `PostHog` instance (never null). The `?.` is misleading — if you want a guard for when telemetry is disabled, add a proper check at the top of `capture()` instead.

**5. Settings GET endpoint leaks `posthog_device_id` — `apps/server/src/routes/settings.ts:11-22`**

The `GET /api/settings` endpoint returns all rows from the settings table, including `posthog_device_id`. This device ID will show up in the frontend settings object. Either filter it out of the GET response or store it in a separate table.

**6. `transcription failed` event includes raw error message — `apps/server/src/routes/transcribe.ts:117`**

```ts
error: err instanceof Error ? err.message : String(err),
```

Error messages from providers can contain API keys, URLs with tokens, or user-adjacent info. Sanitize or just send the error class/name instead of the full message.

**7. Missing telemetry on streaming failures — `apps/server/src/routes/stream.ts:182-201`**

The `postProcess` `.catch()` path in the streaming route captures the Sentry exception but doesn't fire a PostHog event. The non-streaming transcribe route has both `captureException` and `capture("transcription failed", ...)` on error. Keep it consistent.

**8. SIGINT/SIGTERM handlers may swallow existing handlers — `apps/server/src/index.ts:23-24`**

```ts
process.on("SIGINT", () => shutdownPosthog().finally(() => process.exit(0)));
process.on("SIGTERM", () => shutdownPosthog().finally(() => process.exit(0)));
```

These call `process.exit(0)` directly, bypassing any other cleanup registered elsewhere (e.g., Sentry flush, DB close). Either chain with existing shutdown logic or use `process.once` and let the default handler fire after flushing.

### Nit

**9. Empty object literal — `apps/server/src/routes/vocabulary.ts:141`**

```ts
capture("vocabulary term deleted", {});
```

The other events either send meaningful properties or omit the second arg. Pass `undefined` or nothing instead of `{}` for consistency.

MathurAditya724 (changes requested):
## Review (updated)

The instrumentation coverage is solid — events are placed at the right call sites and properties are relevant. Three things need to be addressed before merging.

### 1. No opt-out mechanism — `apps/server/src/lib/posthog.ts`

Telemetry fires unconditionally in all environments. Users have no way to disable it — no setting toggle, no env var, no `DO_NOT_TRACK` header support. For a project whose "privacy is the whole point," shipping always-on analytics without user consent or opt-out is a problem. At minimum:
- Add a `telemetry_enabled` setting (default `true` is fine) and a UI toggle.
- Check it in `capture()` / `captureException()` before sending.
- Respect the [Console Do Not Track](https://consoledonottrack.com/) convention (`DO_NOT_TRACK=1` env var).

### 2. Hardcoded PostHog API key — `posthog.ts:6`

```ts
const POSTHOG_API_KEY = "phc_mDhFafyLK3Safsrrehi7rnH2X9jVMMGNAwKWuJsEN54w";
```

PostHog write keys are designed to be public, so this isn't a secret leak. But hardcoding it means dev/CI builds send events to your production project and you can't rotate without a release. Consider injecting at build time or gating behind `getEnvironment() === "production"` so dev sessions don't pollute your data.

### 3. Remove server-side Sentry — consolidate on PostHog

PostHog already covers exception capture (`captureException`, which this PR already uses) and can handle the metrics currently going through `Sentry.metrics` as event properties. Having two telemetry systems is unnecessary overhead. This PR should also:

- **Delete `apps/server/src/lib/sentry.ts`**
- **Remove `@sentry/node` from `apps/server/package.json`**
- **Replace all server-side Sentry imports** in these files:
  - `apps/server/src/index.ts` — remove `initSentry()` call, replace Sentry `captureException` with PostHog's
  - `apps/server/src/routes/transcribe.ts` — remove `captureException` and `metrics` imports from sentry, use PostHog `capture()` for the count/distribution metrics (you're already capturing `"transcription completed"` / `"transcription failed"` events with `duration_ms`, `audio_duration_ms` etc — those cover the same data)
  - `apps/server/src/routes/stream.ts` — same: remove Sentry `captureException` and `metrics`, the PostHog `"streaming transcription completed"` event already carries `duration_ms` and `audio_duration_ms`
  - `apps/server/src/lib/post-process.ts` — replace Sentry `captureException` with PostHog's, replace `metrics.count("post_process.llm_error")` and `metrics.distribution("post_process.latency")` with a PostHog event like `capture("post process completed", { duration_ms, ... })` / `capture("post process failed", ...)`

The Electron-side Sentry (`@sentry/electron` in `apps/electron/`) is a separate concern and can stay for now — it covers the main process and renderer, which this PR doesn't touch.

### 4. `require()` inside an ESM module — `posthog.ts:31`

```ts
const { getDb } = require("./db.js") as typeof import("./db.js");
```

Every other file in the codebase uses ESM `import`. This `require()` will break if the build moves to pure ESM. Use a dynamic `import()` or restructure so `getDeviceId()` receives the db as a parameter.

MathurAditya724 (commented):
Pushed `58a0585` addressing findings 1-4 from my review:

**What changed:**

1. **Telemetry opt-out** — `capture()` and `captureException()` now check three gates before sending:
   - `DO_NOT_TRACK=1` env var (Console Do Not Track convention)
   - `FREESTYLE_ENV !== "production"` — dev builds no longer send events
   - `telemetry_enabled` setting in the DB — users can disable via the settings API

2. **Removed server-side Sentry** — deleted `sentry.ts`, removed `@sentry/node` from `package.json`, replaced all `captureException` and `metrics.*` calls with PostHog equivalents. The `"transcription completed"`, `"streaming transcription completed"`, and `"post process completed/failed"` events already carry `duration_ms`, `audio_duration_ms`, token counts, etc., so no data is lost.

3. **Fixed `require()` in ESM** — `posthog.ts` now uses a top-level `import { getDb } from "./db.js"` like every other file.

**Not touched:** Electron-side Sentry (`@sentry/electron` in `apps/electron/`) — separate concern, out of scope for this PR.

MathurAditya724 (commented):
All findings from my earlier reviews have been addressed in the commits I pushed:

- **Finding 1 (opt-out)** → `isEnabled()` gate in `posthog.ts` checks `DO_NOT_TRACK`, `FREESTYLE_ENV`, and `telemetry_enabled` setting
- **Finding 2 (dev gating)** → Telemetry only fires when `FREESTYLE_ENV === "production"`
- **Finding 3 (Sentry removal)** → `sentry.ts` deleted, `@sentry/node` removed, all calls replaced with PostHog
- **Finding 4 (require in ESM)** → Replaced with top-level `import`
- **Electron NODE_ENV fix** → `define` in `electron.vite.config.ts` so Sentry actually works in packaged builds
- **skipKeys** → `telemetry_enabled` added alongside `posthog_device_id`

Previous request-changes reviews are now stale.

PR #131, opened 2026-06-01
Fix hotkey binding

### LINKED ISSUE #128: Odd behavior for hotkey binding
As pointed out on Discord:
Freestyle has very odd hotkey binding. It won't let you save control+option as a hotkey but it will let you use only fn as a hotkey, it lets you use just space as a hotkey, and more.

Planned way to address this would be to:
Require a modifier in hotkey binding, and also allow for mouse buttons to be used.
Additionally, allow any combination of modifiers for hotkey.

More on mouse buttons - Many mice have side buttons which allow for them to be easily accessible with the hand that's on them. Left click, right click, middle button are not going to be allowed to be used as a hotkey for obvious reasons. 

I can work on this

tigerisaac:
Closes #128

Changed hotkey rebinding so that it saves intuitively upon release, accepts all modifier combinations and mouse macro buttons. Also it ensures that things like "space" or "enter" won't be accepted as a keybind — it requires modifiers or mouse macro buttons to be included.

Tested on mac platform

### DIFF
--- a/apps/electron/native/linux-key-listener.c
+++ b/apps/electron/native/linux-key-listener.c
@@ -93,6 +93,8 @@ static int parse_key_name(const char *name) {
     if (strcasecmp(name, "CapsLock") == 0) return KEY_CAPSLOCK;
     if (strcasecmp(name, "Pause") == 0) return KEY_PAUSE;
     if (strcasecmp(name, "Insert") == 0) return KEY_INSERT;
+    if (strcasecmp(name, "MouseButton4") == 0 || strcasecmp(name, "Mouse4") == 0) return BTN_SIDE;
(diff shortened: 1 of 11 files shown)


MathurAditya724:
Looks great! Thanks for this

MathurAditya724 (commented):
Nice work — this is a solid improvement to the hotkey binding flow. A couple of things to address before merging:

1. **No way to cancel without Escape** — The Cancel button was removed from the recording UI. The only way to exit recording without saving is pressing Escape, but there's no visible hint telling the user this. Could we add a small "Press Esc to cancel" label in the recording state, or bring back a Cancel button?

2. **Auto-close the issue** — The PR body says "Addresses #128" which won't auto-close the issue on merge. Could you update it to "Closes #128" so it gets resolved automatically?

PR #136, opened 2026-06-01
fix: bundle whisper.cpp binaries in packaged app and fix Windows binary detection

MathurAditya724:
Fixes the "whisper.cpp binary not found" error that occurs in packaged Electron builds. Three root causes addressed:

1. **CI never bundled whisper binaries** — `electron-builder.yml` references `resources/whisper/{platform}-{arch}` in `extraResources`, but CI had no step to populate that directory. Added build/download steps for all three platform jobs.
2. **Windows `X_OK` check is unreliable** — `accessSync(path, X_OK)` doesn't work on Windows (no Unix execute bits). Switched to `F_OK` on Windows so `findExecutable` doesn't miss valid binaries.
3. **Windows zip `Release/` nesting** — the dev download script didn't flatten the upstream zip's `Release/` subdirectory (the runtime `models.ts` had this fix but the dev script didn't). Also added `--resources` flag so CI can place binaries where electron-builder expects them.
4. **Better error surfacing** — `ensureBinariesDownloaded()` failures now propagate with the actual error message instead of falling through to a generic "binary not found".

## Testing
- Server unit tests pass (36/36)
- Biome lint passes on all changed files

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,11 @@ jobs:
         run: pnpm --filter @freestyle/server build
 
       - name: Install native build dependencies
-        run: sudo apt-get update && sudo apt-get install -y libx11-dev libxtst-dev libxext-dev libglib2.0-dev
+        run: sudo apt-get update && sudo apt-get install -y libx11-dev libxtst-dev libxext-dev libglib2.0-dev cmake
+
+      - name: Build whisper.cpp binaries
+        working-directory: apps/electron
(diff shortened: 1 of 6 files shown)


MathurAditya724:
fix-ci: attempt 1 — macOS build fails because whisper.cpp v1.7.5 has a compile error on the CI runner's Apple Silicon (`vmmlaq_s32` requires i8mm target feature). Upgrading to v1.8.5 which fixed this upstream.

PR #139, opened 2026-06-02
Fix MLX ASR runtime release publishing

tigerisaac:
Fixed the workflow to release the MLX worker. 
Works on my mac locally.


matteo8p (approved):
Ship it

PR #141, opened 2026-06-02
feat: add output mode setting to copy instead of paste

### LINKED ISSUE #132: Allow copying transcribed text to clipboard without pasting
Currently, when a transcription finishes, Freestyle always auto-pastes the text into the focused app by simulating `Ctrl+V` / `Cmd+V`. The original clipboard contents are then restored, so the transcribed text is never kept on the clipboard.

It would be useful to have a setting that lets the user **copy the transcribed text to the clipboard only**, without auto-pasting it. This way the user can paste it manually wherever and whenever they want.

## Use case

- Sometimes I want to review the transcribed text before pasting it
- I may want to paste it into multiple places
- Some apps don't play well with simulated paste keystrokes

## Proposed behavior

Add an **Output mode** setting (e.g. in Settings > General) with two options:

| Mode | Behavior |
|---|---|
| **Paste into app** (default, current) | Writes text to clipboard, simulates paste keystroke, restores old clipboard |
| **Copy to clipboard** | Writes text to clipboard and leaves it there — no keystroke simulation |

## Relevant code

- Paste logic: `apps/electron/src/main/paste.ts` → `pasteIntoFocusedApp()`
- IPC handler: `apps/electron/src/main/index.ts` (line ~701) → `ipcMain.handle("paste:text", ...)`
- Renderer call site: `apps/electron/src/renderer/src/pages/app.tsx` (line ~246) → `window.api.pasteText(finalText)`

## Implementation sketch

1. Add a `output_mode` key (`"paste"` | `"clipboard"`) to the settings DB
2. Expose it via IPC + preload API
3. In `app.tsx`, check the setting before calling `pasteText()` — if `"clipboard"`, just write to clipboard via `navigator.clipboard.writeText()` and skip the paste IPC call
4. Add a toggle in the Settings UI

Olaiwonismail:
# Description
Closes #132 
Adds an "Output mode" setting that allows users to choose whether transcribed text is automatically pasted into the active application or copied to the clipboard instead.

This is especially useful for workflows where the user:
- Wants to review transcribed text before pasting it.
- Needs to paste the text into multiple places.
- Uses applications that don't play well with simulated paste keystrokes.

## Changes Made
- **Main Process:** Added a new `copy:text` IPC handler that uses Electron's native `clipboard.writeText()` API.
- **Preload API:** Exposed `window.api.copyText()` to the renderer processes.
- **Frontend (App Logic):** Updated the `drainQueue` transcription handler to fetch the `output_mode` preference from the settings database and route the text to the clipboard if configured.
- **Settings UI:** Added a segment toggle in the **Recording** settings panel for users to switch between "Paste into app" (default) and "Copy to clipboard".

image

## Testing Instructions
1. Navigate to **Settings > General > Recording**.
2. Locate the new **Output mode** toggle.
3. Switch it to **Copy to clipboard**.
4. Record a short clip.
5. Verify that after transcription finishes, the text is not pasted into your active window.
6. Press `Ctrl+V` (or `Cmd+V`) and verify the text was successfully copied to your system clipboard.

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -705,7 +705,13 @@ app.whenReady().then(async () => {
 
   // IPC: copy text to clipboard
   ipcMain.handle("copy:text", async (_event, text: string) => {
+    if (!text?.trim()) return;
     clipboard.writeText(text);
+  });
+
+  // IPC: broadcast output mode changes to pill window
+  ipcMain.on("settings:output-mode-changed", (_event, mode: string) => {
+    mainWindow?.webContents.send("settings:output-mode-changed", mode);
   });
 
   // IPC: hide the pill window on request from renderer
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -28,6 +28,7 @@ import { serve } from "@hono/node-server";
 import {
   app,
   BrowserWindow,
+  clipboard,
   dialog,
   globalShortcut,
   ipcMain,
@@ -700,6 +701,11 @@ app.whenReady().then(async () => {
   // IPC: paste text at cursor
   ipcMain.handle("paste:text", async (_event, text: string) => {
     await pasteIntoFocusedApp(text);
+  });
+
+  // IPC: copy text to clipboard
+  ipcMain.handle("copy:text", async (_event, text: string) => {
+    clipboard.writeText(text);
   });
 
   // IPC: hide the pill window on request from renderer
--- a/apps/electron/src/preload/index.d.ts
+++ b/apps/electron/src/preload/index.d.ts
@@ -5,6 +5,7 @@ declare global {
     electron: ElectronAPI;
     api: {
       pasteText: (text: string) => Promise<void>;
+      copyText: (text: string) => Promise<void>;
       updateHotkey: (hotkey: string) => void;
       reloadHotkey: () => void;
       setHotkeyMode: (mode: "hold" | "toggle") => void;
--- a/apps/electron/src/preload/index.d.ts
+++ b/apps/electron/src/preload/index.d.ts
@@ -66,6 +66,9 @@ declare global {
       onPillPositionChanged: (
         callback: (position: string) => void,
       ) => () => void;
+      // Output mode
+      sendOutputModeChanged: (mode: string) => void;
+      onOutputModeChanged: (callback: (mode: string) => void) => () => void;
       // Hotkey error notifications
       onHotkeyError: (
         callback: (error: { message: string }) => void,
--- a/apps/electron/src/preload/index.ts
+++ b/apps/electron/src/preload/index.ts
@@ -143,6 +143,15 @@ const api = {
     return () =>
       ipcRenderer.removeListener("settings:pill-position-changed", handler);
   },
+  // Output mode
+  sendOutputModeChanged: (mode: string): void =>
+    ipcRenderer.send("settings:output-mode-changed", mode),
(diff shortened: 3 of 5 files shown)


MathurAditya724 (changes requested):
## Review

The PR correctly resolves #132 and follows existing codebase conventions. Clean, minimal diff (55 additions, 2 deletions), CI is green, and the architecture choice of using Electron's main-process `clipboard.writeText()` via IPC is better than the `navigator.clipboard` approach the issue suggested.

Two things worth addressing before merge:

### 1. Per-transcription settings fetch adds latency (medium)

`app.tsx:246-251` — `drainQueue` fetches `output_mode` from the settings API on every single transcription completion. This is a local HTTP round-trip on the critical path, right when the user expects instant paste/copy. The existing `pasteIntoFocusedApp` already has platform-specific settle delays (150-600ms); adding a fetch before that makes it worse.

Other settings that affect runtime behavior are read once on mount or cached. This should do the same — read the setting on mount into a ref and update it when it changes, rather than fetching on every drain.

### 2. Missing empty-string guard on `copy:text` handler (low)

`main/index.ts:707-709` — The existing `pasteIntoFocusedApp` guards with `if (!text?.trim()) return;` at the top. The new `copy:text` handler should do the same to avoid writing empty strings to the clipboard:

```typescript
ipcMain.handle("copy:text", async (_event, text: string) => {
  if (!text?.trim()) return;
  clipboard.writeText(text);
});
```

---

Everything else looks good — defaults are correct (`"paste"` fallback), type declarations are in sync, and the settings UI follows existing patterns exactly.

Olaiwonismail:
fixes and optimizations made to address this:

Removed Fetch Latency: Instead of fetching the output_mode setting from the local database on every single transcription, we now load it into memory once when the pill window mounts.

Added Live Sync: We set up an IPC broadcast (settings:output-mode-changed) so that if the user toggles the setting in the preferences window, it instantly updates the in-memory variable in the pill window without requiring an app restart.

Empty-String Guard: Added defensive programming (if (!text?.trim()) return;) to the copy:text IPC handler to prevent accidentally wiping the user's clipboard history with empty strings if a transcription fails.

> ## Review
> The PR correctly resolves #132 and follows existing codebase conventions. Clean, minimal diff (55 additions, 2 deletions), CI is green, and the architecture choice of using Electron's main-process `clipboard.writeText()` via IPC is better than the `navigator.clipboard` approach the issue suggested.
> 
> Two things worth addressing before merge:
> 
> ### 1. Per-transcription settings fetch adds latency (medium)
> `app.tsx:246-251` — `drainQueue` fetches `output_mode` from the settings API on every single transcription completion. This is a local HTTP round-trip on the critical path, right when the user expects instant paste/copy. The existing `pasteIntoFocusedApp` already has platform-specific settle delays (150-600ms); adding a fetch before that makes it worse.
> 
> Other settings that affect runtime behavior are read once on mount or cached. This should do the same — read the setting on mount into a ref and update it when it changes, rather than fetching on every drain.
> 
> ### 2. Missing empty-string guard on `copy:text` handler (low)
> `main/index.ts:707-709` — The existing `pasteIntoFocusedApp` guards with `if (!text?.trim()) return;` at the top. The new `copy:text` handler should do the same to avoid writing empty strings to the clipboard:
> 
> ```ts
> ipcMain.handle("copy:text", async (_event, text: string) => {
>   if (!text?.trim()) return;
>   clipboard.writeText(text);
> });
> ```
> 
> Everything else looks good — defaults are correct (`"paste"` fallback), type declarations are in sync, and the settings UI follows existing patterns exactly.

MathurAditya724 (approved):
LGTM 🚀

PR #143, opened 2026-06-02
MLX Streaming + Parallel loading

tigerisaac:
MLX will have this speedup a bit before Windows since testing is easier.

Basically, what we're doing is we load the model while the person talks, so that there's no extra latency. Additionally, we reinstate the streaming function for the local models, which was kind of messed up from some of the changes before.

Fun fact, this was written with Freestyle. 

Tested with cloud models and local models, this change doesn't affect cloud models but does speed up local MLX tremendously on shorter generations.

### DIFF
--- a/.github/workflows/mlx-asr-runtime.yml
+++ b/.github/workflows/mlx-asr-runtime.yml
@@ -14,7 +14,7 @@ on:
         description: Runtime release tag to create
         required: true
         type: string
-        default: mlx-asr-worker-v1
+        default: mlx-asr-worker-v2
 
 permissions:
   contents: write
@@ -47,10 +47,10 @@ jobs:
           ASSET: dist/mlx_asr_worker-darwin-arm64.tar.gz
         run: |
           if gh release view "$RELEASE_TAG" >/dev/null 2>&1; then
-            echo "::error::Release $RELEASE_TAG already exists. Use a new runtime tag and update DEFAULT_MLX_WORKER_URL."
+            echo "::error::Release $RELEASE_TAG already exists. Use a new runtime tag and update MLX_WORKER_RELEASE in runtime.ts."
             exit 1
           fi
 
           gh release create "$RELEASE_TAG" "$ASSET" \
             --title "MLX ASR Worker ${RELEASE_TAG#mlx-asr-worker-}" \
-            --notes "Optional Freestyle runtime for local Qwen3 ASR."
+            --notes "Optional Freestyle runtime for local MLX ASR (Qwen3, Parakeet, Whisper, etc.)."
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -20,7 +20,6 @@ import { join } from "node:path";
 import { pathToFileURL } from "node:url";
 import { electronApp, is, optimizer } from "@electron-toolkit/utils";
 import server, {
-  autoStartMlxAsrServer,
   autoStartWhisperServer,
   reconcileUnsupportedMlxVoiceDefault,
 } from "@freestyle/server";
@@ -829,7 +828,6 @@ app.whenReady().then(async () => {
   // Run non-critical server startup tasks now that the DB path is set
   reconcileUnsupportedMlxVoiceDefault();
   autoStartWhisperServer();
-  autoStartMlxAsrServer();
 
   // Start the Hono HTTP server with WebSocket support (or reuse an existing one)
   function startServer(port: number): void {
--- a/apps/electron/src/renderer/src/lib/models.ts
+++ b/apps/electron/src/renderer/src/lib/models.ts
@@ -217,6 +217,7 @@ export const LOCAL_VOICE_NOTES: Record<string, string> = {
   "qwen3-0.6b-5bit": "Fast · great for low memory",
   "qwen3-0.6b-8bit": "Balanced quality and size",
   "qwen3-1.7b-8bit": "Best Qwen accuracy",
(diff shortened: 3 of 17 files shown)


MathurAditya724 (commented):
## Review

The core idea — loading the MLX worker in parallel with audio capture — is sound, and the `workerReadyPromise` pattern in `mlx-local.ts` is well-designed. A few issues to address before merging:

---

### 1. `session.ready` sent before cloud providers are actually ready (High)

Removing the client-side `sessionReady` gate means audio is now forwarded to the upstream provider as soon as the WebSocket is open. For MLX this is fine (audio accumulates in `chunks[]` and inference awaits `workerReadyPromise`). But for cloud providers (Deepgram, ElevenLabs, OpenAI), their `waitUntilReady()` returns `Promise.resolve()` immediately, so `afterSessionReady` fires `notifySessionReady` → `flushPendingAudio()` before the cloud provider's own WebSocket has opened.

The old flow was: provider WS opens → `onReady` fires → `flushPendingAudio` + send `session.ready`. The new flow is: `waitUntilReady` resolves instantly → flush + ready sent → audio forwarded to a not-yet-open provider socket. Audio in that window gets dropped.

Fix options:
- Make cloud providers' `waitUntilReady` resolve only when their internal WS is actually open, or
- Keep the `onReady`-driven flow for cloud providers and only use `afterSessionReady` for MLX.

### 2. `onReady` callback is gutted but still called by all providers (Medium)

`onReady` in `stream.ts` was stripped of its two main jobs (flushing audio + sending `session.ready`) and now only resets `reconnectAttempts`. But all providers still call it. For cloud providers, `onReady` fires when their WS opens — that's the real "ready" signal. Now the server ignores it and instead relies on `waitUntilReady()` (which resolves instantly for cloud). This creates a timing mismatch.

Either keep `onReady` as the trigger for cloud providers, or remove it from the interface entirely and make all providers use `waitUntilReady` as the sole readiness signal.

### 3. `waitUntilReady()` boilerplate in cloud providers is unnecessary (Low)

The three identical additions to Deepgram, ElevenLabs, and OpenAI:
```ts
waitUntilReady(): Promise<void> {
  return Promise.resolve();
}
```
are not needed. `waitUntilReady` is already optional on the `StreamSession` interface, and `afterSessionReady` handles the absent case with `session.waitUntilReady?.() ?? Promise.resolve()`. These can be removed.

### 4. `try/catch` in preconnect `useEffect` is dead code (Low)

```tsx
useEffect(() => {
  try {
    getStreamer();
  } catch {
    // Server may not be up yet; startCapture will retry.
  }
}, [getStreamer]);
```

`getStreamer()` is synchronous — it constructs the `Streamer` which calls `openWebSocket()` synchronously. Server-down errors are async (they fire on the WS `error`/`close` events, not in the constructor). The `try/catch` would only catch a synchronous throw from `new WebSocket(url)` (e.g., invalid URL). The comment is misleading; this catch block is almost certainly dead code.

### 5. `onReady` parameter mismatch (Nit)

The `onReady` callback signature changed from `(model) => { ... }` to `() => { ... }`, but MLX still passes the model ID at the call site. Harmless at runtime but inconsistent.

MathurAditya724 (commented):
## Follow-up Review

All five original findings have been addressed.

| # | Finding | Status |
|---|---------|--------|
| 1 | `session.ready` timing for cloud providers | **Fixed** — `afterSessionReady` now returns early when `waitUntilReady` is absent. Cloud providers flow through `onReady` → `notifySessionReady` as before. Audio buffering on the server is correctly gated: `(!upstream.waitUntilReady && notifiedReadyToken !== readyToken)` queues audio until `onReady` fires. |
| 2 | `onReady` gutted but still called | **Fixed** — `onReady` now delegates to `notifySessionReady`, which handles flushing + sending `session.ready`. The `readyToken`/`notifiedReadyToken` pair makes it idempotent, so for MLX where both `onReady` and `afterSessionReady` can resolve, whichever fires first wins. |
| 3 | Unnecessary `waitUntilReady` boilerplate | **Fixed** — removed from all three cloud providers. No diff to those files. |
| 4 | Dead `try/catch` in preconnect `useEffect` | **Fixed** — the preconnect `useEffect` was removed entirely. The streamer is now created lazily on first `getStreamer()` call in `startRecording`. |
| 5 | `onReady` parameter mismatch | **Fixed** — callback now receives `(readyModel)` and passes it through. |

A few observations on the new code (none are blockers):

**`commit()` preview heuristic in `mlx-local.ts`** — When no partial has been emitted yet (`!lastText`) and enough audio is buffered, `commit()` runs a non-final inference first before the final one. This adds an extra inference round for short recordings. Seems intentional for UX (show something before the final result). Worth noting this doubles inference cost for quick recordings.

**`stopMlxServer` shutdown change** — The `SIGTERM` was replaced with `stdin.end()` after writing the shutdown JSON. Shutdown now relies entirely on the Python process reading the stdin message. The 5-second `SIGKILL` timeout is the only fallback. Fine as long as the Python worker reads stdin promptly.

**`autoStartMlxAsrServer` removal** — Consistent with the lazy/parallel loading design. The MLX server is now started on first streaming session rather than at app launch.

**`onConfig` in `app.tsx`** — New logic: `if (wantsMicRef.current) { sessionStreamingRef.current = config.streaming; }`. This handles the case where the config message arrives after `startRecording` has already set `sessionStreamingRef`. Good defensive fix.

The `readyToken`/`notifiedReadyToken` mechanism is a clean solution that correctly handles both cloud and MLX readiness paths. LGTM.

PR #149, opened 2026-06-02
Fix transcription language in streaming path + macOS Accessibility prompt/URL

BransTiong:
Two independent bug fixes, both verified locally (`pnpm biome check .`, `pnpm --filter @freestyle/electron typecheck`, and `pnpm --filter @freestyle/server test` → 36/36 pass).

## 1. Streaming transcription ignores the Language setting

**Symptom:** With a specific language chosen under *Settings → Recording → Language*, dictation still sometimes comes out in the wrong language.

**Cause:** Only the batch `POST /transcribe` path read the `language` setting. The streaming `/stream` path called `openStreamingSession()` without it, so streaming-capable models (OpenAI realtime, Deepgram, ElevenLabs, local MLX) always auto-detected.

**Fix:** Thread the existing `language` setting through `StreamingSessionOptions` into each streaming provider's upstream request:
- OpenAI realtime → `input_audio_transcription.language`
- Deepgram → `language` query param
- ElevenLabs → `language_code` query param
- MLX local worker → `language`

An unset/`"auto"` value preserves auto-detection, so default behavior is unchanged.

## 2. macOS: Accessibility "Open Settings" goes to the wrong pane and never prompts

**Symptom:** The in-app *Open Settings* button opens the **Accessibility features** pane (VoiceOver/Zoom/…) instead of **Privacy & Security → Accessibility**, and the app never appears in the list on its own.

**Cause:** It used the legacy `x-apple.systempreferences:com.apple.preference.security?Privacy_Accessibility` URL (lands on the wrong pane on recent macOS) and only ever called `isTrustedAccessibilityClient(false)` (never prompts).

**Fix:** Call `isTrustedAccessibilityClient(true)` to fire the native "control your computer" prompt (which auto-adds the app to the list), and open the modern `com.apple.settings.PrivacySecurity.extension?Privacy_Accessibility` URL.

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -765,10 +765,15 @@ app.whenReady().then(async () => {
     return true;
   });
 
-  ipcMain.on("permissions:open-accessibility", () => {
+  ipcMain.on("permissions:open-accessibility", async () => {
     if (process.platform === "darwin") {
+      // Passing `true` pops the native "would like to control this computer"
+      // prompt and adds Freestyle to the Accessibility list automatically, so
+      // the user only has to flip the toggle (macOS never lets us flip it).
+      const { systemPreferences } = await import("electron");
+      systemPreferences.isTrustedAccessibilityClient(true);
       shell.openExternal(
-        "x-apple.systempreferences:com.apple.preference.security?Privacy_Accessibility",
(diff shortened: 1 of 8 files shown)


MathurAditya724 (approved):
LGTM 🚀 , thank you for this

PR #152, opened 2026-06-03
Implement "Hard Reset" that completely resets Freestyle (Dev only) 

matteo8p:
# Overview

In this PR we implement hard reset, a way to completely reset and wipe out Freestyle. This feature is only available on dev. We want to add this feature to enable testing for onboarding, View what onboarding is like from a clean user slate.


MathurAditya724:
@matteo8p I have previously added a menu option for "Reset Onboarding", should we merge this with that?

matteo8p:
@MathurAditya724 The reset onboarding feature is already merged in. This is a separate feature, doing an entire hard reset and wipe. I think we should have both, both, serve their own purpose. Onboarding reset is to just quickly view, the onboarding pages. But I also want a hard reset option, to fully simulate a new user experience end to end. 

Is that what you are asking about?

MathurAditya724 (commented):
Clean, well-scoped PR. The dev-only gating is correct, the dialog UX is clear, the file deletion list is appropriate (keeps models, removes config + DB), and the shutdown sequence is thorough. CI passes clean.

One item worth discussing:

**`app.relaunch()` + `app.exit(0)` runs outside the try/catch** (`index.ts:585-586`). If cleanup partially fails (e.g. `rm` throws on one file), the catch logs the error but the app still relaunches into a potentially half-reset state (DB deleted but `settings.json` still present, or vice versa). Moving `app.relaunch()` / `app.exit(0)` inside the try block would prevent relaunch on failure — or if best-effort-then-relaunch is intentional, a comment would help future readers. For a dev-only tool this is pragmatic either way.

matteo8p:
@MathurAditya724 Addressed! All operations are in a try catch block and we surface any failures if they happen, as a dialog

PR #154, opened 2026-06-03
fix: prevent MLX ASR worker unload while transcriptions are in flight

MathurAditya724:
Fixes a race condition where `scheduleUnload` could stop the MLX ASR worker process while transcription requests were still pending, causing `mlx-asr worker stopped` errors during active transcription.

The idle-unload timer and `scheduleUnload` did not check whether requests were in flight before killing the worker. When a transcription was active and the unload timer fired (or `scheduleUnload` was called from a streaming session's cancel/close), `stopMlxServer` would terminate the process and reject all pending requests.

## Changes

Three guards added to `apps/server/src/lib/mlx-asr/server.ts`:

- **`scheduleUnload`** — skip entirely when `pending.size > 0`
- **Unload timer callback** — re-check `pending.size` before stopping (a new request may have arrived between scheduling and firing)
- **`sendTranscribeRequest`** — cancel any pending unload timer before adding the request to the `pending` map

## Testing

- `pnpm build` passes (server + Electron typecheck and bundle)
- `pnpm test` — all 36 server API tests pass

### DIFF
--- a/apps/server/src/lib/mlx-asr/server.ts
+++ b/apps/server/src/lib/mlx-asr/server.ts
@@ -404,6 +404,8 @@ function sendTranscribeRequest(opts: {
   stream?: boolean;
   onPartial?: (text: string) => void;
 }): Promise<string> {
+  clearUnloadTimer();
+
   const proc = workerProcess;
   if (!proc?.stdin || !workerReady) {
     return Promise.reject(new Error("mlx-asr worker is not running"));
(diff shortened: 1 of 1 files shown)


MathurAditya724:
All 9 CI checks pass (Lint, Typecheck, Server API, Electron E2E, Linux/Windows/macOS builds, CI Status). Self-review confirmed every call path through `scheduleUnload`, `sendTranscribeRequest`, and `applyMlxAsrRetentionPolicy` is correctly guarded — no edge cases where pending requests could be killed or where the unload timer leaks without being rescheduled.

PR #155, opened 2026-06-03
fix: register Escape cancel shortcut on all platforms, not just Windows

MathurAditya724:
The Escape key cancellation during transcription was only registered on Windows via a `process.platform === "win32"` guard. On macOS and Linux, pressing Escape during recording/transcription did nothing — the pill stayed visible and transcription continued.

This removes the platform guard so the `globalShortcut.register("Escape", ...)` call in `showPill()` and the matching `globalShortcut.unregister("Escape")` in `hidePill()` run on all platforms.

## Testing
Manual — verify Escape cancels recording/transcription on macOS, Windows, and Linux.

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -871,9 +871,7 @@ app.whenReady().then(async () => {
       keyListener.stop();
       keyListener = null;
     }
-    if (process.platform === "win32") {
-      globalShortcut.unregisterAll();
-    }
+    globalShortcut.unregisterAll();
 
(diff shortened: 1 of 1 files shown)


MathurAditya724:
PR is ready for review. CI passed (Preview Changelog check). Self-review caught three additional `globalShortcut.unregisterAll()` sites that were still guarded by `process.platform === "win32"` — those have been fixed in the second commit so the Escape shortcut doesn't leak on macOS/Linux during hotkey re-registration, hotkey recording, or app quit.

PR #157, opened 2026-06-03
fix: update download button not triggering download or showing progress

### LINKED ISSUE #156: [Bug] "Download" button on update modal (v0.1.0 -> v0.1.2) does nothing when app is launched via CLI
When the update prompt appears in the UI (as seen in Screenshot 2026-06-03 at 11.31.20.jpg), clicking the blue Download button fails to trigger any network activity, download progress, or external browser event.

Image

MathurAditya724:
The "Download" button in the native update dialog (tray menu → "Check for Updates...") called `autoUpdater.downloadUpdate()` without updating `updateDownloadState` or notifying the settings window via IPC. This caused two problems:

1. **No state tracking**: `updateDownloadState` stayed `"idle"`, so if the user opened the settings page during a download, the banner showed "Download" instead of "Downloading...".
2. **Unhandled promise rejection**: The Promise returned by `downloadUpdate()` was never caught, so download failures could surface as unhandled rejections instead of being routed through the `autoUpdater` error event.

The fix sets `updateDownloadState = "downloading"` and sends `updater:downloading` to the settings window before starting the download (matching what the IPC `updater:download` handler already does), and catches the returned Promise in both code paths.

## Testing
Manual — trigger "Check for Updates..." from the tray menu, click "Download", then open Settings to verify the banner shows "Downloading..." state. Also verify download errors surface in the settings banner.

Closes #156

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -588,6 +588,15 @@ function restartAndUpdate(): void {
   autoUpdater.quitAndInstall();
 }
 
+/** Mark state as downloading, notify the settings window, and kick off the download. */
+function triggerDownloadUpdate(): void {
+  updateDownloadState = "downloading";
+  settingsWindow?.webContents.send("updater:downloading");
+  autoUpdater.downloadUpdate().catch((err) => {
+    log.warn(`downloadUpdate rejected: ${err}`);
+  });
+}
+
 async function checkForUpdatesFromMenu(): Promise<void> {
   if (is.dev) {
     dialog.showMessageBox({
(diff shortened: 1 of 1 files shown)


MathurAditya724:
PR is ready for review. CI passed (Preview Changelog). Self-review found a DRY issue: the download-triggering pattern was duplicated between the menu dialog path and the IPC handler with subtle differences (the IPC handler wasn't sending `updater:downloading` to the settings window, and the original menu path had no error catching). Extracted a `triggerDownloadUpdate()` helper that both sites now use consistently, with error logging in the catch.

PR #158, opened 2026-06-03
feat: validate API keys on save with per-provider health checks

MathurAditya724:
Validate API keys when users add or edit them by hitting lightweight, free provider endpoints (list-models, list-projects, etc.) before saving. Keys with known bad prefixes (e.g. non-`sk-` for OpenAI) are rejected immediately without a network call.

The validation status is persisted in the DB and surfaced in the UI with check/X icons in the providers list, a spinner during validation, and inline error banners in both the settings and onboarding key dialogs.

## Testing
- `pnpm run -r build` passes (typecheck + production build)
- `pnpm run test` passes (36/36 tests)

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -588,6 +588,15 @@ function restartAndUpdate(): void {
   autoUpdater.quitAndInstall();
 }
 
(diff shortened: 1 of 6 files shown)


MathurAditya724:
Superseded by new PR based on main — the original branch included unrelated changes from a stale base.

PR #164, opened 2026-06-03
Enhance 111: Adding panning to pill widget

srirae:
I have fixed the panning issue when playing around with the widget. I've also taken care about the changes in the settings tab. The pill location is automatically saved and can be manually tinkered in the settings. The key changes are in the render sections where the `webkit` has been changed from no-drag to drag so the pill can be moved around.

I've also added a video for the enhancement PR

https://github.com/user-attachments/assets/86ffb6c3-6243-4210-b551-f51f3dcb46c8

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -175,18 +175,40 @@ function getDashboardURL(path = "/"): string {
   return `app://renderer${path}`;
 }
 
-let isProgrammaticMove = false;
+// Tracks the exact coordinates of the last programmatic setPosition call.
+// The move listener compares reported coords against this target and ignores
+// matching events, eliminating the fixed-timeout race condition.
+let programmaticTarget: { x: number; y: number } | null = null;
+let programmaticCleanupTimer: NodeJS.Timeout | null = null;
 
 function setProgrammaticPosition(
   win: BrowserWindow,
   x: number,
   y: number,
 ): void {
-  isProgrammaticMove = true;
+  programmaticTarget = { x, y };
+  // Safety: clear the target after 1s in case the OS never delivers a settle event.
+  if (programmaticCleanupTimer) clearTimeout(programmaticCleanupTimer);
+  programmaticCleanupTimer = setTimeout(() => {
+    programmaticTarget = null;
+    programmaticCleanupTimer = null;
+  }, 1000);
   win.setPosition(x, y);
-  setTimeout(() => {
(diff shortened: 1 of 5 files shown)


MathurAditya724 (changes requested):
## Review

The feature (drag-to-reposition the pill) is useful and the approach is reasonable — listen for `move` events, debounce, persist to "custom" position. A few issues:

### Bug: race between `setTimeout` flag reset and `move` events

`isProgrammaticMove` is reset via `setTimeout` with a fixed delay (200ms for `setProgrammaticPosition`, 500ms for `createAppWindow`). If the OS delivers `move` events after the timeout fires but before the programmatic move fully settles — or if the system is under load — a programmatic reposition will be saved as a "custom" position, overwriting the user's chosen preset.

This is inherently fragile. A safer pattern: set a flag *before* calling `setPosition`, and clear it in the *next* `move` event callback (or after a single tick), rather than guessing a timeout.

### Issue: `WebkitAppRegion: "drag"` on the entire pill blocks click interactions

The diff changes the pill's inner style from `no-drag` to `drag`:

```ts
WebkitAppRegion: "drag",
```

`-webkit-app-region: drag` makes the element act as a window title bar — it captures all mouse events for window dragging. This means click/mousedown/mouseup events on the pill (buttons, text, interactive elements inside `pillInnerStyle`) will no longer fire. Users won't be able to interact with anything inside the pill; they can only drag it.

If the pill has interactive children (cancel button, status text, etc.), those need `WebkitAppRegion: "no-drag"` explicitly. Otherwise this breaks all pill interactions on every platform, not just the drag feature.

### Nit: `window.screenY` check for custom position alignment is unreliable

```ts
pos === "custom"
  ? window.screenY < window.screen.availHeight / 2
  : !!pos?.startsWith("top")
```

`window.screenY` is the BrowserWindow's screen coordinate, which depends on monitor arrangement and may be negative on multi-monitor setups. Comparing against `screen.availHeight / 2` assumes a single monitor starting at y=0. On a vertically stacked dual-monitor setup, the pill could be at the top of the lower monitor but have a large `screenY` value.

### Nit: IIFE in JSX for `positionOptions`

The `(() => { ... })()` pattern in the settings page JSX is unusual. A simpler approach: compute `positionOptions` as a `const` before the return, or use `useMemo`.

srirae:
Hi, I've made the requested changes.

MathurAditya724:
Thank you for this PR

PR #166, opened 2026-06-03
Modify our onboarding flow.

matteo8p:
# Overview.

In this PR we introduce a new onboarding flow. We've simplified the number of steps in the onboarding flow.
 Now to three steps.

1. Configure your computer settings, accessibility permissions, and mic permissions.
2. Configure your model. we decided to have Qwen be the default model if MLX is compatible or fallback to whisper.
3. Test to your setup in the text input box.

Screenshot 2026-06-03 at 8 18 56 AM

Screenshot 2026-06-03 at 8 19 04 AM

Screenshot 2026-06-03 at 8 23 18 AM


apps/electron/src/renderer/src/components/voice-row.tsx
```diff
@@ -180,7 +183,7 @@ export function VoiceRow({
                   <TooltipContent>Download size</TooltipContent>
                 </Tooltip>
               )}
-              {item.ram && (
+              {!hideRam && item.ram && (
```
matteo8p: TODO: Remove

PR #167, opened 2026-06-03
fix: wait for native key listener READY before confirming hotkey registration

### LINKED ISSUE #165: Can't start recording on Ubuntu
I'm using the latest release of Ubuntu 26.04. I tried Freestyle with .deb and AppImage, but I can't trigger the recording. I set the shortcut to different alt+key shortcuts. And when I'm back to the home screen, nothing happens when I press the shortcut.

Here the log when I start the deb from console:

`~$ freestyle
(node:40448) ExperimentalWarning: SQLite is an experimental feature and might change at any time
(Use `freestyle --trace-warnings ...` to show where the warning was created)
13:40:31.656 info [electron] Reusing existing Freestyle server on http://localhost:4649
13:40:31.818 info [updater] Checking for update
[40484:0603/134031.842409:ERROR:media/gpu/vaapi/vaapi_wrapper.cc:1631] vaInitialize failed: unknown libva error
13:40:33.270 info [updater] Update for version 0.1.3 is not available (latest version: 0.1.3, downgrade is disallowed).
13:40:33.643 info [whisper] Server ready on port 8178
[40448:0603/134035.677437:ERROR:base/nix/mime_util_xdg.cc:139] Invalid mime.cache file does not contain null prior to ALIAS_LIST_OFFSET=44
[40448:0603/134035.677475:ERROR:base/nix/mime_util_xdg.cc:139] Invalid mime.cache file does not contain null prior to ALIAS_LIST_OFFSET=44
`

MathurAditya724:
On Linux, when the user lacks `/dev/input` permissions, the native key listener binary exits immediately but `start()` already returned `true` — so the `globalShortcut` fallback was never activated and the hotkey silently stopped working.

`start()` now returns a `Promise<boolean>` that resolves `true` only after the binary emits `READY`, or `false` if it exits first. `registerHotkey()` awaits the result and falls back to `globalShortcut` when the native listener fails. stderr is always logged so the `/dev/input` permission hint reaches users. `globalShortcut` cleanup is no longer gated on Windows only.

## Testing
`npm run typecheck` passes (both `:node` and `:web`). Manual testing requires an Ubuntu machine without `input` group membership — the fix ensures the app falls back to `globalShortcut` and shows a descriptive error if both mechanisms fail.

Closes #165

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -1417,7 +1417,7 @@ function handleNativeHotkeyUp(): void {
   }
 }
 
-function registerHotkey(hotkey?: string): void {
+async function registerHotkey(hotkey?: string): Promise<void> {
   // Tear down previous listener
   if (keyListener) {
     keyListener.stop();
@@ -1436,11 +1436,13 @@ function registerHotkey(hotkey?: string): void {
   currentHotkeyAccel = accel;
 
   // Try native key listener binary first (all platforms)
+  let nativeError = "";
   keyListener = new NativeKeyListener({
     hotkey: accel,
     onKeyDown: handleNativeHotkeyDown,
     onKeyUp: handleNativeHotkeyUp,
     onError: (error) => {
+      nativeError = error;
       hotkeyLog.error(`Native key listener error: ${error}`);
     },
     onReady: () => {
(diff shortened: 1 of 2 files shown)


MathurAditya724:
CI passed (Preview Changelog check). Self-review found and fixed two additional issues: a race condition where rapid `registerHotkey` calls could overlap and corrupt each other's state, and a leftover `win32`-only gate on `globalShortcut.unregisterAll()` in the hotkey recorder handler. Both typechecks (`:node` and `:web`) pass. Ready for human review.

PR #169, opened 2026-06-03
Feat: auto-update managed mlx worker

tigerisaac:
Summary:
Build mlx_asr_worker-darwin-arm64.tar.gz on ever macOS CI run and publish via Craft alongside the app. We remove the standalone workflow.
Download/cache the frozen worker to ~/.cache/freestyle/mlx-asr/runtime/ when users don’t have dev Python + mlx-audio.
Ties worker update to actual app updates: when electron-updater finishes downloading a new version, stage the worker for that release in the background; on first launch after restart, promote the staged worker (does not replace the live worker while the old app is still running).
First MLX model download or worker start still calls updateManagedMlxRuntimeIfNeeded() if staging did not run (e.g. manual DMG install).
Worker fingerprint (MLX_WORKER_BUILD_SPEC + mlx_asr_server.py hash) avoids redundant downloads when the app semver bumps but the worker build is unchanged.

Important notes:
On every macOS release, Build (macOS) must pass (./scripts/build_mlx_asr_worker.sh). Craft release must include mlx_asr_worker-darwin-arm64.tar.gz (.craft.yml includeNames).
When someone changes the worker (deps or scripts/mlx_asr_server.py), they must keep scripts/build_mlx_asr_worker.sh and MLX_WORKER_BUILD_SPEC in runtime.ts in sync.


MathurAditya724 (changes requested):
## Review

Well-structured approach to tying the MLX worker lifecycle to app updates. The staging/promote pattern, fingerprint-based skip logic, and opt-in gating (only users with downloaded MLX models) are all sound design choices. Tests cover the key versioning scenarios well.

One critical bug and a couple of secondary issues:

### Bug: `promoteStagedRuntime` deletes the staged content before renaming it

`stagedRuntimeRoot(releaseTag)` returns `join(getMlxRuntimeDir(), "staging", releaseTag)` — a **subdirectory** of `getMlxRuntimeDir()`. In `promoteStagedRuntime` (runtime.ts:340-355):

```ts
rmSync(runtimeDir, { recursive: true, force: true });  // line 347 — deletes staging/ too
mkdirSync(dirname(runtimeDir), { recursive: true });
renameSync(stagedRoot, runtimeDir);                     // line 349 — ENOENT: stagedRoot is gone
```

The `rmSync` on line 347 recursively deletes the entire `runtimeDir`, which includes `runtimeDir/staging/releaseTag`. The subsequent `renameSync` will always fail because the source path no longer exists.

**Fix**: Store the staging directory outside the runtime dir (e.g., sibling: `join(getMlxCacheDir(), "staging", releaseTag)`) so it survives the `rmSync`.

### Nit: dead `tempDir` variable in `downloadRuntime`

Lines 414/416 compute the same string twice:
```ts
await downloadRuntimeToDir(active, `${runtimeDir}.downloading`, url);   // line 414
const tempDir = `${runtimeDir}.downloading`;                             // line 416
```
`tempDir` could be declared once before the `downloadRuntimeToDir` call. Minor readability issue, no functional impact.

### Test gap: no test for `promoteStagedRuntime` / `activateManagedMlxRuntimeForAppVersion` with actual staged files

The existing test "does not prefetch or activate the runtime until an MLX model is downloaded" only checks the early-return guard. There's no test that sets up a staged runtime on disk and verifies `promoteStagedRuntime` actually works (which would also catch the bug above).

MathurAditya724 (approved):

PR #175, opened 2026-06-04
Electron paste issue

matteo8p:
Fix #1 — repair the osascript fallback (electron-builder.yml)

  Added the missing NSAppleEventsUsageDescription to extendInfo. Under the notarized hardened runtime,
  sending Apple Events to System Events is blocked without this Info.plist string — which is why the
  osascript Cmd+V fallback silently failed in prod. With it present, macOS shows the Automation prompt
  and the fallback works once approved.

  Fix #2 — stop masking the missing Accessibility permission (src/main/index.ts)

  - Removed accessibilityConfirmed = true from the globalShortcut fallback branch (was line 1682).
  Registering a global shortcut needs no Accessibility permission, so latching the flag there produced a
  false positive that hid the "grant Accessibility" prompt and left the native fast-paste path broken in
  prod.

  - The flag is now set only when the native key listener starts (real proof of Accessibility).
  
  - Updated the three stale comments (flag declaration, the registered branch, and the
  check-accessibility handler) to match.

  Why this fixes prod

  - Your dev Electron binary was granted Accessibility long ago, so both paste layers worked locally. The
  notarized prod app is a fresh code identity that had been granted nothing, and notarize: true enforces
  TCC strictly.
  - After #1, paste works via the osascript fallback as soon as the user approves Automation.
  - After #2, the app correctly reports Accessibility as not granted, so onboarding will prompt for it —
  enabling the fast native CGEvent path (macos-fast-paste) instead of the slower fallback.

### DIFF
--- a/apps/electron/electron-builder.yml
+++ b/apps/electron/electron-builder.yml
@@ -37,6 +37,7 @@ mac:
     NSMicrophoneUsageDescription: Application requests access to the device's microphone.
     NSDocumentsFolderUsageDescription: Application requests access to the user's Documents folder.
     NSDownloadsFolderUsageDescription: Application requests access to the user's Downloads folder.
+    NSAppleEventsUsageDescription: Freestyle uses Apple Events to paste transcribed text into the focused application.
   notarize: true
   extraResources:
     - from: "resources/whisper/darwin-${arch}"
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -745,7 +745,9 @@ async function factoryReset(): Promise<void> {
     app.relaunch();
     app.exit(0);
   } catch (err) {
-    console.error("[factory-reset] failed:", err);
+    log.error(
+      `factory-reset failed: ${err instanceof Error ? err.message : String(err)}`,
+    );
     dialog.showErrorBox(
       "Hard Reset failed",
       `${err instanceof Error ? err.message : String(err)}\n\nThe app may be in a partially reset state. Quit and relaunch manually.`,
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -130,9 +130,12 @@ let mainWindow: BrowserWindow | null = null;
 let settingsWindow: BrowserWindow | null = null;
 let tray: Tray | null = null;
 let keyListener: NativeKeyListener | null = null;
-// Latching flag: once the native key listener (or globalShortcut fallback)
-// has started successfully, accessibility is confirmed. The flag persists
-// even when keyListener is temporarily torn down for hotkey recording.
+// Latching flag: set only once the native key listener has started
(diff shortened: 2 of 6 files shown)


MathurAditya724 (changes requested):
## Review

This PR bundles two targeted paste/accessibility fixes with ~20 unrelated PRs worth of changes (56 files, +3674/−1801). The stated intent — adding `NSAppleEventsUsageDescription` and fixing the `accessibilityConfirmed` false positive — accounts for roughly 7 lines of the diff. Everything else (logger migration, onboarding rewrite, MLX runtime staging, analytics overhaul, pill panning, API key validation, CI changes, etc.) is scope that makes this very difficult to review or bisect if something breaks. I'd strongly recommend splitting this into separate PRs per concern.

That said, here are the issues I found in the actual code:

### Bugs

**1. `updateHotkey()` orphans pending `start()` promise** — `key-listener.ts:495`

`updateHotkey()` calls `proc.removeAllListeners()` on the old process before calling `this.start()`. If the previous `start()` call is still awaiting its READY signal, the promise will **never resolve** because the `close`/`error` handlers that would settle it are stripped. The caller in `index.ts` (`const started = await listener.start()`) would hang forever.

**2. `console.error` missed in logger migration** — `index.ts:748`

The `factoryReset` catch block still uses `console.error("[factory-reset] failed:", err)` while every other call in the file uses the structured `log.*` logger. Should be `log.error(...)`.

**3. `readInstalledRuntimeMetadata()` reads from destroyed directory** — `runtime.ts:430`

After `rmSync(runtimeDir)` on line 426 and renaming the temp dir into its place on line 428, `readInstalledRuntimeMetadata()` reads from the *new* directory which has no `metadata.json` yet. The old `syncedAppVersion` is always lost. The metadata should be read *before* the `rmSync`.

**4. Concurrent stdout interleaving in Python progress reporter** — `mlx_asr_server.py:83-118`

`_ProgressTqdm` declares a `_lock` but never acquires it. Up to 8 download threads can call `_emit()` → `_send()` → `print()` concurrently, which can interleave output and produce malformed JSON on stdout. The Node caller parses these as newline-delimited JSON, so corrupted lines will cause parse errors.

### Style

**5. Import/declaration ordering in `whisper.ts`** — `whisper.ts:7-9`

`const log = createAppLogger("whisper")` sits between two import blocks. Move it after all imports.

### Scope

**6. PR bundles ~20 unrelated features/fixes** 

The core fix is 2 files / ~7 lines. The rest includes: new `packages/utils` package, full onboarding rewrite, analytics migration (PostHog client-side → server-side), API key validation system, MLX runtime auto-update lifecycle, pill drag/panning, tray menu Linux fix, streaming language passthrough, CI runner + build changes, CHANGELOG additions for 0.1.3 and 0.1.4, and more. These should be individual PRs.

MathurAditya724 (approved):

PR #189, opened 2026-06-05
Use Huggingface Hub for model downloads

matteo8p:
# Overview

Right now when we are downloading local models we are doing a fetch method directly. This is less reliable as we have to write all of our custom error handlers. This is what `@huggingface/hub` is supposed to handle for us, that download reliability. 

In this PR we replaced our fetch logic with methods from `@huggingface/hub`.

### DIFF
--- a/docs/specs/mlx-hub-download-migration.md
+++ b/docs/specs/mlx-hub-download-migration.md
@@ -1,0 +1,245 @@
+# Technical Spec: Migrate MLX ASR Downloads to `@huggingface/hub`
+
+**Status:** Draft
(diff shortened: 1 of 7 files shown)


docs/specs/mlx-hub-download-migration.md
```diff
@@ -0,0 +1,245 @@
+# Technical Spec: Migrate MLX ASR Downloads to `@huggingface/hub`
```
matteo8p, line 1: Ignore this .md file. This is future work, migrating MLX to also use `@huggingface/hub`

PR #190, opened 2026-06-05
Use `@huggingface/hub` to download MLX models

matteo8p:
# Overview

In this PR, we use `@huggingface/hub` to download MLX models. Keep in mind that MLX models are only available on Mac OS.

This is a follow-up PR to https://github.com/freestyle-voice/freestyle/pull/189 where we replace our custom fetch logic with `@huggingface/hub` as well. I also removed some code comments around the surface area.

### DIFF
--- a/apps/server/src/lib/mlx-asr/models.ts
+++ b/apps/server/src/lib/mlx-asr/models.ts
@@ -1,22 +1,16 @@
-import { type ChildProcess, spawn } from "node:child_process";
 import { existsSync, readdirSync, rmSync } from "node:fs";
 import { homedir } from "node:os";
 import { join } from "node:path";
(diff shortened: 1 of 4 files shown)


apps/server/src/lib/mlx-asr/models.ts
```diff
@@ -272,78 +235,33 @@ export async function downloadMlxModel(modelId: string): Promise<void> {
...
-        resolve();
-        return;
-      }
-
-      active.error =
-        active.stderr.trim() ||
-        `MLX model download failed with exit code ${code ?? "unknown"}`;
-      reject(new Error(active.error));
+  try {
+    await snapshotDownload({
```
matteo8p, line 251: This is the main code block to look out for.

PR #208, opened 2026-06-07
fix: percent-encode x-app-context header to avoid fetch failure on non-Latin1 window titles

vbp1:
## Problem

Dictating into some apps (e.g. Notepad++) fails with `failed to execute fetch` and no text is inserted.

The pill sends the focused window context in the `x-app-context` HTTP header of `POST /api/transcribe`. On Windows that value is JSON containing the window title, which for editors like Notepad++ includes the full path of the open file. When the path contains characters outside ISO-8859-1 (for example a Cyrillic file name), the browser's `fetch()` throws synchronously:

```
TypeError: Failed to execute 'fetch' on 'Window': Invalid value
```

because HTTP header values are restricted to Latin-1. Apps whose window title is plain ASCII are unaffected, which is why the failure only shows up on some apps.

## Fix

- Client (`app.tsx`): percent-encode the `x-app-context` value before putting it in the header (both the main transcribe path and the post-stream fallback).
- Server (`transcribe.ts`): decode the value back with `decodeURIComponent`, falling back to the raw string for compatibility with older clients.

The decoded JSON is passed to `postProcess` unchanged, so context-aware formatting keeps working.

## Testing

- `biome check` clean on the changed files
- `typecheck` passes for both server and electron
- Server test suite green (pre-existing macOS-only `mlx-runtime` failures are unrelated and also fail on `main` in my Windows environment)

### DIFF
--- a/apps/electron/src/renderer/src/pages/app.tsx
+++ b/apps/electron/src/renderer/src/pages/app.tsx
@@ -106,6 +106,17 @@ interface TranscribeResult {
   error?: string;
 }
 
+/**
+ * The app context (process name + window title) can contain characters
+ * outside ISO-8859-1 — e.g. a Cyrillic file path in the Notepad++ title
+ * bar. HTTP header values only allow Latin-1, so passing the raw JSON
+ * makes fetch() throw "Failed to execute 'fetch'". Percent-encode it so
+ * the header is always byte-safe; the server decodes it back.
+ */
+function encodeAppContext(context: string): string {
+  return encodeURIComponent(context);
+}
+
 interface QueueEntry {
   promise: Promise<TranscribeResult>;
 }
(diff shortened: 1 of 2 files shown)


MathurAditya724 (commented):
Clean fix for a real bug. The approach — percent-encode on the client, decode on the server — is the right way to handle non-Latin1 characters in HTTP headers.

Checked:
- Both `x-app-context` header sites in `app.tsx` (streaming path and post-stream fallback) are covered.
- The JSON-body path (`/api/post-process`) doesn't pass through headers, so correctly doesn't need encoding.
- `decodeAppContext` catches decode errors and falls back to the raw string, preserving backward compat with older clients.
- CI green (lint, typecheck, tests, e2e, Windows build all pass; Linux/macOS builds still pending but unrelated to this change).

LGTM — no issues found.

Issue #188, opened 2026-06-05
Button "launch at startup" doesn't stay on


Upacesky-26:
I'm running Ubuntu 26.04 and installed Freestyle 0.1.5 via deb.
When I want to autostart Freestyle, I activate the "Launch at startup" button, and it seems activated at first, but if I change the view and come back to settings, the button is deactivated.
I expect it to stay activated and start the app at the beginning of the session.

matteo8p:
Got it. Thank you so much for reporting this issue. Let me take a look at it and get back to you.

PR #210, opened 2026-06-07
fix: prevent multiple instances on Linux and persist autostart setting

### LINKED ISSUE #191: Can start the app multiple times
On Ubtuntu 26.04, if I close freestyle with the x, the app goes to tray, so far so good, but if I start Freestyle again from my menu, it opens as a new app and all the processes start again too.
Expected behaviour, when the app runs, calling it again opens the window from the app that's already started, and doesn't start a new instance.

MathurAditya724:
## Summary

Fixes #191 and #188 — two Ubuntu-specific issues.

- **Multiple instances (#191)**: The app had no single-instance lock. Closing the window (which goes to tray) and relaunching from the menu started a completely new process — duplicate tray icon, server, hotkey listeners. Added `app.requestSingleInstanceLock()` so the second instance quits immediately and the primary instance's window is focused.

- **Launch at startup not persisting (#188)**: `app.setLoginItemSettings()` / `app.getLoginItemSettings()` are macOS + Windows only — on Linux they're no-ops. The toggle appeared to work (React state updated) but reverted on next launch. Added XDG-compliant autostart via a `.desktop` file in `$XDG_CONFIG_HOME/autostart/`, with the IPC handlers and factory reset branching on `process.platform`.

### E2E tests

Added `tests/linux-autostart.test.ts` that runs on `ubuntu-latest` CI:
- Verifies enabling autostart creates a `.desktop` file with correct content
- Verifies `isEnabled()` returns true after enabling
- Verifies disabling removes the file
- Verifies `isEnabled()` returns false after disabling
- Verifies the app holds the single-instance lock

Tests use `XDG_CONFIG_HOME` pointed at a temp directory to avoid side effects.


MathurAditya724:
@Upacesky-26 can you check if this PR solves your issue

Issue #181, opened 2026-06-05
Create the ability to view total cost within time ranges.


matteo8p:
# Overview

Right now, the history page only shows cost all time. However, we don't really want to show cost all time, all the time. We want to let the user choose which time range they want to see the total cost.

In the all time text, have it be underlined. When you click on it, there's a drop down that lets you select today. This week, and all time. Then, when you click on this week, it'll show only the total cost for this week. 

<img width="1134" height="730" alt="Image" src="https://github.com/user-attachments/assets/c1201683-69b0-458a-bd96-3375655374c9" />

# Directions
1. Make the cost card. A hyperlink or a button that you can click on to change the time range.
2. Create an action that changes the time range when you hit "this week" or "today".
3. Recalculate the total cost and render that based off of the time region selected.

# First time contributing? 

If this is your first time contributing, I highly recommend reading our [CONTRIBUTING.md](https://github.com/freestyle-voice/freestyle/blob/main/CONTRIBUTING.md). It has instructions on how to set up. 

Also consider joining our discord server. This is where contributors communicate. We have all project discussions there. Happy to provide assistance there, too.

<p align="left">
  <a href="https://discord.gg/Fmgt5yZCDu"><img src="https://img.shields.io/badge/Discord-Join%20Server-5865F2.svg?style=for-the-badge&logo=discord&logoColor=white" alt="Discord" /></a>
</p>

matteo8p:
If you're interested in this issue, please drop a comment, and I will assign the issue to you. First come, first serve basis.

shawnriju:
@matteo8p Hey Matt, this issue looks pretty interesting. I'd be happy to take it on and submit a fix if no one else is working on it.
(Also, just FYI, Raidou_14th is my username on Discord)

matteo8p:
@shawnriju Would love to have you take on this issue, assigning to you. Sickk let's chat there

matteo8p:
@shawnriju Let's also have the default selection be "weekly"

Agastya18:
status??

andrewjsalzer:
Does this issue need to be closed?

matteo8p:
@andrewjsalzer Ahh good callout. Yes this needs to be closed out. 

@Agastya18 We finished this issue!

PR #211, opened 2026-06-07
Feat/history filter panel

shawnriju:
## Summary

This PR addresses #181 to implement a general-purpose date range filter panel for the transcription history page.

Following feedback from @MathurAditya724 , I skipped the cost-card dropdown and implemented a proper slide-in filter panel drawer. This allows users to set custom date ranges to filter both the list and the metrics, that includes showing the transcription costs as per the filtered data.

### Backend Updates (history.ts)
- GET `/api/history`: Added optional `start_date` and `end_date` parameters. SQL queries dynamically append WHERE conditions comparing `date(created_at, 'localtime')` to filter the items.
- GET `/api/history/stats`: Added optional `start_date` and `end_date` support to aggregate transcription stats (sessions, total words, average latency, and costs) dynamically for the active range.

### Frontend UI & Integration(history.tsx)
- **Default Weekly View:** Set default view of the history page to the past 7 days (Weekly preset) instead of showing all-time records.
- **Filter Drawer:** Added a slide-in panel from the right, providing Start Date and End Date calendar inputs, quick presets (Today, Last 7 Days, Last 30 Days, All Time), and a footer action layout containing "Clear All" and "Apply Filters".
- **Interactive Highlights:** Added highlighting for the active-preset buttons by dynamically matching active dates to preset ranges.
- **Clamping Date Constraints:** Added validation guards on both inputs (min on End Date, max on Start Date, and automatic handler clamping) to prevent setting a Start Date that exceeds the End Date.
- **UX Fix (Empty States):** Split the empty states into two separate components:
  - EmptyState (onboarding guide): Retained only for the initial fresh-user state when the database is completely empty and no filters are set.
  - NoSearchResults: A new, dedicated empty state designed specifically for cases when filters or search queries yield 0 items. This preserves the search and filter controls on the screen and features a "Clear filters and search" action link.
- **Stats Cards:** Changed the stat labels from static "all time" labels to dynamic "words filtered" and "cost · filtered" whenever a date filter is applied.

## Manual Verification

- **Default State:** Loaded the history page and verified it automatically filters to the past 7 days.
- **Preset Toggles:** Clicked presets ("Today", "Last 30 Days", "All Time") and verified that buttons highlight correctly and active date bounds align.
- **Date Constraints:** Set a Start Date past the selected End Date and verified that inputs clamp the date bounds correctly.
- **Empty State Reset:** Filtered to a future date range with 0 entries, verified the filter/search controls remained on-screen, and clicked "Clear filters and search" to restore the list.

### DIFF
--- a/apps/electron/src/renderer/src/pages/history.tsx
+++ b/apps/electron/src/renderer/src/pages/history.tsx
@@ -1,3 +1,9 @@
+import {
+  Sheet,
+  SheetContent,
+  SheetHeader,
+  SheetTitle,
+} from "@renderer/components/ui/sheet";
 import { getClient } from "@renderer/lib/api";
 import { cn } from "@renderer/lib/utils";
 import {
@@ -6,6 +12,7 @@ import {
   ChevronRight,
   Clock,
   Copy,
+  Filter,
   Search,
   Trash2,
 } from "lucide-react";
@@ -65,6 +72,13 @@ function formatCost(cost: number): string {
   return `$${cost.toFixed(3)}`;
 }
 
+function getLocalDateString(d: Date): string {
+  const year = d.getFullYear();
+  const month = String(d.getMonth() + 1).padStart(2, "0");
+  const day = String(d.getDate()).padStart(2, "0");
+  return `${year}-${month}-${day}`;
+}
+
 /** Get a date key for grouping: "Today", "Yesterday", or "Day, Mon DD" */
 function getDateGroup(iso: string): string {
   const d = new Date(`${iso}Z`);
@@ -92,6 +106,15 @@ export default function HistoryPage(): React.JSX.Element {
   const [search, setSearch] = useState("");
   const [stats, setStats] = useState<Stats | null>(null);
   const [loading, setLoading] = useState(true);
+  const [startDate, setStartDate] = useState(() => {
+    const d = new Date();
+    d.setDate(d.getDate() - 7);
+    return getLocalDateString(d);
+  });
+  const [endDate, setEndDate] = useState(() => {
+    return getLocalDateString(new Date());
+  });
+  const [filterOpen, setFilterOpen] = useState(false);
 
   const loadData = useCallback(async () => {
     try {
@@ -101,11 +124,17 @@ export default function HistoryPage(): React.JSX.Element {
         orderBy: "-created_at",
       };
       if (search) query.search = search;
+      if (startDate) query.start_date = startDate;
+      if (endDate) query.end_date = endDate;
+
+      const statsQuery: Record<string, string> = {};
+      if (startDate) statsQuery.start_date = startDate;
+      if (endDate) statsQuery.end_date = endDate;
 
       const client = getClient();
       const [histRes, statsRes] = await Promise.all([
         client.api.history.$get({ query }),
-        client.api.history.stats.$get(),
+        client.api.history.stats.$get({ query: statsQuery }),
       ]);
       if (histRes.ok) {
         const data = await histRes.json();
@@ -118,7 +147,7 @@ export default function HistoryPage(): React.JSX.Element {
     } finally {
       setLoading(false);
     }
-  }, [page, search]);
+  }, [page, search, startDate, endDate]);
 
   useEffect(() => {
     loadData();
@@ -166,7 +195,20 @@ export default function HistoryPage(): React.JSX.Element {
     );
   }
 
-  const isEmpty = total === 0 && !search;
+  const todayStr = getLocalDateString(new Date());
+  const start7 = new Date();
+  start7.setDate(start7.getDate() - 7);
+  const start7Str = getLocalDateString(start7);
+  const start30 = new Date();
+  start30.setDate(start30.getDate() - 30);
+  const start30Str = getLocalDateString(start30);
+
+  const isTodayPreset = startDate === todayStr && endDate === todayStr;
+  const isWeeklyPreset = startDate === start7Str && endDate === todayStr;
+  const isMonthlyPreset = startDate === start30Str && endDate === todayStr;
+  const isAllTimePreset = startDate === "" && endDate === "";
+
+  const isGenuineEmpty = total === 0 && !search && !startDate && !endDate;
 
   return (
     <div
@@ -180,7 +222,7 @@ export default function HistoryPage(): React.JSX.Element {
       >
         <PageHeader title="History" />
 
-        {isEmpty ? (
+        {isGenuineEmpty ? (
           <EmptyState />
         ) : (
           <>
@@ -188,7 +230,7 @@ export default function HistoryPage(): React.JSX.Element {
             <div className="border-border mb-7 grid grid-cols-2 gap-2.5 border-b pb-7 md:grid-cols-4">
               <Stat
                 n={(stats?.total_words ?? 0).toLocaleString()}
-                l="words all time"
+                l={startDate || endDate ? "words filtered" : "words all time"}
               />
               <Stat n={String(stats?.total_sessions ?? 0)} l="sessions" />
               <Stat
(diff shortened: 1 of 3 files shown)


MathurAditya724 (changes requested):
## Review

Thanks for the work on this! The filter panel approach is solid and the dynamic WHERE builder in the backend is a nice improvement. I found a few issues worth addressing before merging.

---

### Bugs

**1. "Apply Filters" button is misleading — filters apply immediately, not on click**

The "Apply Filters" button only does `setFilterOpen(false)` (closes the drawer). But `startDate`/`endDate` state updates immediately when the user interacts with date inputs or preset buttons, which triggers the `useEffect` → `loadData` chain on every change. The data is already fetched before the user clicks "Apply Filters."

This creates confusing UX: users expect "Apply" means "submit my choices now." Either defer state changes to local/draft state until "Apply" is clicked, or remove the "Apply" button and just auto-close/auto-apply (which is what actually happens).

**2. First-time users with zero transcriptions will never see the onboarding `EmptyState`**

`isGenuineEmpty = total === 0 && !search && !startDate && !endDate`. Since `startDate` is initialized to 7 days ago (never empty on first load), `isGenuineEmpty` will always be `false` — even for a brand-new user with zero transcriptions. They'll see the stats cards (all zeros) and `NoSearchResults` ("no transcripts found for this date range") instead of the onboarding guide. The onboarding `EmptyState` is effectively unreachable.

A fix would be to track whether the database has *any* transcriptions at all (e.g., a separate unfiltered count, or a flag from the API) and use that to decide between `EmptyState` vs `NoSearchResults`.

**3. Preset date computations can desync from initial state after midnight**

`todayStr`, `start7Str`, `start30Str` are computed on every render via `new Date()`. If the user leaves the page open past midnight, the preset highlighting will update (the computed strings move forward a day), but `startDate`/`endDate` state still holds the old values from mount time. The "Last 7 Days" button will stop being highlighted even though the user hasn't changed anything. Not a crash, but a subtle inconsistency.

**4. No server-side date format validation**

`start_date` and `end_date` query params are passed directly to SQLite with no format validation. While they are properly parameterized (no injection risk), a non-date string like `"foo"` would cause `date(created_at, 'localtime') >= 'foo'` to do a lexicographic comparison and return unexpected results. A simple regex check (`/^\d{4}-\d{2}-\d{2}$/`) on the server would harden this.

---

### Design / UX

**5. Default changed from "All Time" to "Last 7 Days"**

This is a significant behavior change — a user who transcribed something 8 days ago won't see it on first load. The PR description mentions this intentionally, but it's worth confirming this is the desired behavior with the maintainers.

**6. Filter badge shows a checkmark `✓` instead of a count**

The filter button shows a `✓` when filters are active. Most filter UIs show the count of active filters (e.g., "2" for start + end date). Minor, but worth considering.

---

### Code Quality

**7. Comment style: missing space after `//`**

```typescript
//Query rows
//Query total count
```

Should be `// Query rows` and `// Query total count` per the project's style.

**8. No tests added**

The PR adds date filtering to both the list and stats endpoints but includes no tests. The existing test coverage for history is minimal (only empty-state tests). Adding at least a basic test that inserts a row and verifies `start_date`/`end_date` filtering works would be valuable.

**9. `date(created_at, 'localtime')` prevents index usage**

This function is applied to every row during WHERE evaluation, preventing SQLite from using an index on `created_at`. For typical history table sizes this is fine (and matches the existing `today_sessions` query pattern), but worth noting if the table grows large. A more performant approach would be converting user-supplied local dates to UTC bounds on the server side.

**10. Dynamic WHERE builder is a good pattern**

The `conditions[]` / `params[]` approach is cleaner than the existing if/else branching in `dictionary.ts`, `vocabulary.ts`, etc. Not a blocker, but the new pattern creates an inconsistency with those routes. Worth adopting more broadly if the maintainers like it.

shawnriju:
@MathurAditya724 I've made the changes based on the above feedback. Let me know if everything looks good now.

MathurAditya724 (commented):
## Follow-up Review

Nice work addressing the previous feedback — the `activePreset` refactor, `unfiltered_total_sessions`, date validation, "Done" button rename, and the new tests are all solid improvements. A few remaining issues:

---

### New Bug: `NoSearchResults` shows wrong message on default preset

When the user is on the default "weekly" preset (never touched any filters) and searches for something that matches nothing, `hasDates={!!(startDate || endDate)}` is `true` because `startDate`/`endDate` are always non-empty for the weekly preset.

This means the message reads *"no transcripts match that search and date range"* instead of just *"no transcripts match that search"*. The user never set a date range — they're on the default. The "Clear filters and search" button also appears, but clicking it resets to "weekly" (already the current preset), so only the search text actually changes.

**Fix:** Use `hasDates={activePreset !== "weekly"}` (or `activePreset === "custom" || activePreset === "today" || ...`) to only treat non-default presets as "has date filters."

### UX nit: `filterCount` badge semantics

```typescript
const filterCount = activePreset !== "weekly" ? 1 : 0;
```

"All Time" (no date restriction) shows a badge of "1", while "Weekly" (actively restricts to 7 days) shows no badge. This is technically "deviation from default = badge" but can confuse users who interpret the badge as "data is being narrowed." Consider whether showing a badge for "All Time" is the right call — it's the least restrictive option.

### Nit: `DATE_REGEX` should be hoisted to module scope

`const DATE_REGEX = /^\d{4}-\d{2}-\d{2}$/` is declared inside both route handlers, allocating a new regex on every request. Move it next to `ALLOWED_ORDER_COLUMNS` at module level.

### Test timezone fragility

The test's `formatDate` uses local timezone methods (`getFullYear`, `getMonth`, `getDate`) to build datetime strings that are inserted into `created_at` (which the server treats as UTC via `date(created_at, 'localtime')`). In timezones with large positive UTC offsets, `date("2026-06-08 12:00:00", 'localtime')` can produce `"2026-06-09"`, causing the test's date filter assertions to fail. This likely works fine in CI (typically UTC) but could flake locally. Using `getUTCFullYear()`/`getUTCMonth()`/`getUTCDate()` would make the test timezone-safe.

---

The `NoSearchResults` message bug is the main actionable item. The rest are minor nits. Overall looking good — the preset-based architecture is much cleaner than the previous version.

apps/electron/src/renderer/src/pages/history.tsx
```diff
@@ -202,30 +273,58 @@ export default function HistoryPage(): React.JSX.Element {
...
+                  </span>
+                )}
+              </button>
             </div>
 
             {entries.length === 0 ? (
-              <NoSearchResults />
+              <NoSearchResults
+                hasSearch={!!search}
+                hasDates={!!(startDate || endDate)}
```
MathurAditya724: **Bug:** `hasDates` is always `true` on the default "weekly" preset because `startDate`/`endDate` are non-empty (`start7Str`/`todayStr`). When a user on the default preset searches and gets no results, the message reads *"no transcripts match that search **and date range**"* — but they never set a date range.

The "Clear filters and search" button also appears, but clicking it resets to "weekly" (the same preset they're already on) — only the search text actually clears, which is confusing.

**Suggested fix:** Use `hasDates={activePreset !== "weekly"}` so the default preset is not treated as an active date filter.

apps/electron/src/renderer/src/pages/history.tsx
```diff
@@ -92,6 +107,53 @@ export default function HistoryPage(): React.JSX.Element {
...
+  const getTimeLabel = (): string => {
+    if (activePreset === "weekly") return "past 7 days";
+    if (activePreset === "today") return "today";
+    if (activePreset === "monthly") return "past 30 days";
+    if (activePreset === "all-time") return "all time";
+    return "filtered"; // custom range
+  };
+  const timeLabel = getTimeLabel();
+
+  const filterCount = activePreset !== "weekly" ? 1 : 0;
```
MathurAditya724: **Nit:** "All Time" (no date restriction at all) shows a badge of `1`, while "Weekly" (actively restricts to 7 days) shows `0`. This is semantically inverted — the least restrictive option shows a filter badge. Consider whether the badge should mean "data is being narrowed" vs. "deviated from default."

apps/server/src/routes/history.ts
```diff
@@ -27,9 +27,18 @@ const ALLOWED_ORDER_COLUMNS = new Set([
 const history = new Hono()
   .get("/", (c) => {
     const db = getDb();
+    const DATE_REGEX = /^\d{4}-\d{2}-\d{2}$/;
```
MathurAditya724: **Nit:** `DATE_REGEX` is re-created on every request. It's also duplicated in the `/stats` handler (line 97). Hoist it to module scope next to `ALLOWED_ORDER_COLUMNS`:
```ts
const DATE_REGEX = /^\d{4}-\d{2}-\d{2}$/;
```

apps/server/tests/api.test.ts
```diff
@@ -398,4 +404,116 @@ describe("History", () => {
...
+        duration_ms, audio_duration_ms, input_tokens, output_tokens, cost_usd, created_at
+      ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+    `);
+
+    const now = new Date();
+
+    const formatDate = (d: Date, daysAgo = 0) => {
+      const target = new Date(d);
+      target.setDate(target.getDate() - daysAgo);
+      const year = target.getFullYear();
```
MathurAditya724: **Latent flake:** `formatDate` uses local-timezone methods (`getFullYear`, `getMonth`, `getDate`) to build datetime strings inserted into `created_at` (stored/treated as UTC). The server compares with `date(created_at, 'localtime')`.

In timezones with large positive UTC offsets (e.g. UTC+13), `date("2026-06-08 12:00:00", 'localtime')` produces `"2026-06-09"`, causing the row to fall outside the test's expected range.

Using `getUTCFullYear()`/`getUTCMonth()`/`getUTCDate()` would make it timezone-safe. Likely passes in CI (usually UTC) but could fail locally.

MathurAditya724 (changes requested):
The preset-based refactor is a big improvement — nice work addressing the previous feedback. A few remaining issues, one medium bug and a few nits.

shawnriju:
Whoops, yeah I missed that bug. Pushed the fix for that and the other nits!

matteo8p (approved):
I was able to test this out. It works quite well. Thank you so much for putting this in @shawnriju!!

PR #220, opened 2026-06-09
feat: live partial preview in the recording pill

anish808:
Update: Live partial preview is now toggle (hands-free) mode only. Hold mode keeps the waveform bars.

Show streaming STT text in the pill while dictating, including local Whisper via incremental inference. Long partials scroll to keep the newest words visible.


matteo8p:
@anish808 Going to close this out for now, not a feature we want today.

PR #230, opened 2026-06-10
Update platform default push-to-talk hotkeys

tigerisaac:
## Summary
- Set macOS default hotkey to `Fn` (Globe key)
- Set Windows default hotkey to `Control+Super` (Ctrl + Win)
- Set Linux default hotkey to `Alt+Super` (common dictation combo, avoids WM launcher binds)
- Update renderer fallbacks to use the shared `getDefaultHotkey()` helper

## Test plan
- [ ] Fresh install / cleared hotkey setting on macOS defaults to Fn
- [ ] Fresh install on Windows defaults to Control+Super
- [ ] Fresh install on Linux defaults to Alt+Super
- [ ] Settings and onboarding show the correct default before a custom hotkey is saved
- [ ] Existing users with a saved hotkey setting are unchanged


matteo8p (approved):
Love this, seriously think we needed a single source of truth function, and changes in the default. Thanks for putting that in.

Issue #232, opened 2026-06-11
Bug: Pressing Esc to cancel transcription would freeze the running hotkey for the next invocation


RezqZiyad:
Thanks for the awesome app!

After ESC is pressed to cancel a transcription, the running hotkey wouldn't respond on the next invocation and I have to double use the hotkey for it to run.

This is on Windows, running latest release.

anish808:
Good catch @RezqZiyad , noticed it on mac as well. fix is underway

matteo8p:
This should be fixed now. Please try it out and let me know if it works. Going to close out this issue for now.

PR #233, opened 2026-06-11
fix: restore hotkey after ESC cancels dictation (#232)

anish808:
After pressing Escape to cancel, the next hotkey press was ignored until a second press. Sync pill state refs immediately on cancel/hide, ignore stale hotkey-up events when the session is inactive, and reset the main process hotkey latch when the pill closes.


matteo8p (approved):
Ship itttt

Issue #138, opened 2026-06-02
Feature: Add FunASR/SenseVoice as local STT engine


LauraGPT:
Hi! Great voice-to-text dictation app — local first!

Would you consider [FunASR](https://github.com/modelscope/FunASR) or [SenseVoice](https://github.com/FunAudioLLM/SenseVoice) as a local STT engine?

## Why?

- **Ultra-fast** — SenseVoice-Small: ~70ms for 10s audio
- **Built-in punctuation** — Natural dictation output
- **Streaming** — Paraformer-streaming for real-time dictation
- **50+ languages** — Including CJK with state-of-the-art accuracy
- **Fully local** — Aligns with local-first approach
- **Apache 2.0**

## References

- FunASR: https://github.com/modelscope/FunASR (16K+ stars)
- SenseVoice: https://github.com/FunAudioLLM/SenseVoice (8K+ stars)

tigerisaac:
Depends. On Mac, we have Qwen + MLX which accomplishes everything for us already. If FunASR and SenseVoice are available as MLX models it shouldn’t be a problem to add.

matteo8p:
@LauraGPT Hey Zhifu, thanks for putting this issue out, really appreciate it. I'm not familiar with FunASR or SenseVoice but happy to take a look at them and I'm open to adding more models. 

Are you a developer? Would you be willing to contribute these models into Freestyle? Would love to have you as a part of the community.

LauraGPT:
Thanks @matteo8p! Yes, I'm from the FunASR team. Would be happy to contribute.

For MLX (@tigerisaac), SenseVoice's non-autoregressive architecture is well-suited for MLX — in fact, @lucasnewman just submitted a Fun-ASR-Nano MLX implementation at [mlx-audio #760](https://github.com/Blaizzy/mlx-audio/pull/760).

I'll take a look at Freestyle's codebase and see how to best integrate FunASR/SenseVoice. Will open a PR when ready!

matteo8p:
@LauraGPT That is really awesome. I'm happy to have you here, and I would be really grateful for any contributions. Please feel free to take a look at the codebase. I'd be happy to help in any way. I'm thinking that if these models are MLX compatible, it wouldn't be a huge lift to add them into Freestyle. 

Also, a lot of the contributors and I communicate on Discord, so we'd love to have you there too!

tigerisaac:
[...]

PR #236, opened 2026-06-11
Add SenseVoice Small as a Local MLX ASR model

LauraGPT:
## Summary

Adds SenseVoice Small as a Local MLX ASR option and makes the MLX worker more tolerant of result shapes returned by different MLX audio STT models.

Changes:
- add `sensevoice-small` to the Local MLX ASR catalog, backed by `mlx-community/SenseVoiceSmall`
- add a catalog test so the model remains registered
- make `scripts/mlx_asr_server.py` extract text from object, dict, and list result shapes
- add worker tests for dict/list text extraction

This keeps the default provider behavior unchanged; users only see an additional downloadable Local MLX model.

Refs #138.

## Verification

- `python3 scripts/test_mlx_asr_server.py` passed
- `pnpm --filter @freestyle/server test -- apps/server/tests/mlx-models.test.ts` not run locally because this server image has no `node`, `npm`, `pnpm`, or `corepack` on PATH. CI should cover this test.

### DIFF
--- a/apps/server/src/lib/mlx-asr/constants.ts
+++ b/apps/server/src/lib/mlx-asr/constants.ts
@@ -54,6 +54,17 @@ export const MLX_ASR_MODELS: MlxAsrModelDef[] = [
     quantized: true,
   },
   {
+    id: "sensevoice-small",
+    hfId: "mlx-community/SenseVoiceSmall",
+    family: "sensevoice",
+    displayName: "SenseVoice",
+    sizeBytes: 936_000_000,
+    ramRequired: "~1.5 GB",
+    speed: "Fast",
+    quality: "High",
+    quantized: false,
+  },
+  {
     id: "parakeet-tdt-0.6b-v3",
     hfId: "mlx-community/parakeet-tdt-0.6b-v3",
     family: "parakeet",
--- a/apps/server/tests/mlx-models.test.ts
+++ b/apps/server/tests/mlx-models.test.ts
@@ -1,0 +1,15 @@
+import { describe, expect, it } from "vitest";
+import { MLX_ASR_MODELS } from "../src/lib/mlx-asr/constants.js";
+
+describe("MLX ASR model catalog", () => {
+  it("includes SenseVoice Small as a local MLX transcription model", () => {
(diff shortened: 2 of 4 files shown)


tigerisaac:
A few thoughts: on initial tests for Sensevoice Small, its main value is being very fast (which is insane, it's near instant). However, the MLX port doesn't allow for word boosting or punctuation, and the RAM usage is similar to Qwen3 0.6b, and Qwen does word boosting + punctuation.
Was wondering your view on it, what is the benefit of offering Sensevoice Small? We've been trying to better scrutinize models and providers added so that we don't plague the users with a huge amount of models to choose from.
Thanks again for your contribution.

LauraGPT:
Thanks for testing it and for the thoughtful framing.

My view is that SenseVoice Small is valuable as a “fast local fallback / low-latency” option rather than a full replacement for the existing higher-feature model path. The main benefits are:

- Very low latency for short dictation and command-style transcription.
- Strong local/offline privacy story with a small model that feels responsive on Apple Silicon.
- Good multilingual / code-switching behavior for common dictation use cases.
- A useful option for users who care more about speed than word boosting or punctuation quality.

I agree the MLX port limitation matters: no word boosting and weaker/no punctuation means it should not be presented as the best general-purpose default. If you want the model list to stay curated, I’m happy to adjust this PR so SenseVoice is clearly labeled as something like “Fast local, no word boosting” and hide/disable unsupported controls for it. Another option is to keep it behind an “advanced/local models” section rather than putting it on equal footing with the more fully-featured provider.

So the benefit I’m aiming for is not “more accurate than the existing option in every setting,” but “near-instant local transcription for users who prefer speed and offline operation.” Happy to make the UX copy/placement match that positioning.

LauraGPT:
Hi! SenseVoice Small is a great fit for freestyle's MLX ASR support. Happy to help with any adjustments needed! 🙏

matteo8p (approved):
@LauraGPT Thank you so much for putting in this PR! As @tigerisaac said, we are being more scrutinized in what models we provide. Diminished returns on the more models that we have in Freestyle.

We'll be capping it to 1-2 models per model provider. I'm happy to have SenseVoice be a part of that family.

PR #240, opened 2026-06-11
Add Soniox realtime STT with upgraded LLM processing pipeline

tigerisaac:
First impressions of this change are pretty good.
Summary:
Add Soniox (soniox/stt-rt-v4) as a cloud voice provider with WebSocket streaming, manual finalize on commit, and language_hints from the language setting
Wire vocabulary hotwords to Soniox via context.terms
Port the benchmark cleanup prompt into editor/prompts.ts and use it from a single postProcess() path (batch, streaming, and Soniox handoff)
Soniox fast handoff: prewarm + optional pipeline timings; works with any supported cleanup LLM; dictionary replacements on all paths
Dynamic maxOutputTokens scales with input length (floor 512, cap 8192)
Soniox upstream sessions are ephemeral per recording (session-policy) to avoid long-lived billing sessions

Basically just a new provider and upgraded prompt to LLM for post processing. Nothing crazy but performance is pretty good


apps/server/src/lib/editor/model-hints.ts
```diff
@@ -0,0 +1,41 @@
+/** Remove a trailing paragraph duplicated from earlier in the output. */
+export function stripTrailingDuplicate(text: string): string {
```
matteo8p, line 2: Do you know why this is needed? Does Qwen STT really add thinking tags etc?

tigerisaac: It was the LLM post process that kept leaking thinking tags - my bad for including this, it was a testing thing that accidentally passed over. I took it out

matteo8p (approved):
Ship

apps/server/src/lib/editor/model-hints.ts
matteo8p: Ahhh I understand, thanks for catching that.

PR #244, opened 2026-06-12
Enhance: Added new languages for i18n localization

srirae:
Hi,

I have added new languages to the software and will be adding more in the future. I have added a video below for your reference.

[PR #2]: I have structured the entire translation process. There is a proper markdown for all translators to check before adding translations. The file can be found in `apps\electron\src\renderer\src\locales\README.md`. The video below is from the first PR and many things has been changed.

https://github.com/user-attachments/assets/e8f872b6-2a80-4eb8-89ad-7ec20a0f1a48


matteo8p (approved):
Ship it

PR #246, opened 2026-06-12
fix(linux): show app icon in GNOME launcher and dock

### LINKED ISSUE #243: Linux: app icon not shown in GNOME (launcher and dock)
> Disclaimer: this issue was written with the help of AI (Claude), based on a real debugging session on the reporter's machine.

**Setup:** Freestyle 0.1.8 (.deb) on Ubuntu, GNOME on Wayland.

The app shows a generic gear icon everywhere. Two separate packaging causes, both verified locally:

1. **Icon installed at an unregistered size.** The .deb only ships `/usr/share/icons/hicolor/1024x1024/apps/freestyle.png`, but the hicolor theme (`index.theme`, hicolor-icon-theme 0.18) has no `1024x1024` directory — its sizes stop at `512x512`. Icon lookup never searches unregistered directories, so the icon is invisible. Fix: install at standard sizes (e.g. 512x512, 256x256) or as `scalable/apps/freestyle.svg`.

2. **StartupWMClass doesn't match the real window app_id.** `package.json` has `name: "@freestyle/electron"` and no `productName`, so the Wayland app_id is literally `@freestyle/electron` (confirmed via `WAYLAND_DEBUG=1`, the client sends `set_app_id("@freestyle/electron")`). The desktop file ships `StartupWMClass=Freestyle`, so GNOME can't associate the running window with `freestyle.desktop`, and the dock/taskbar falls back to a gear. Fix: set `productName` in package.json, or set `StartupWMClass=@freestyle/electron` in the generated desktop file.

**Workaround used locally:** copy the icon to a registered hicolor size and override the desktop file with the correct `StartupWMClass` — both icons then display correctly.

andrewjsalzer:
## Summary
- Ship Linux icons at standard hicolor sizes (16–512) instead of only 1024x1024, which GNOME does not look up
- Set `desktopName` and `StartupWMClass` so Electron's Wayland app_id matches `freestyle.desktop`

Fixes #243

## Test plan
- [x] Manual workaround on installed 0.1.8 confirmed icons appear in GNOME launcher and dock
- [x] Built `.deb` locally and verified icons install under `/usr/share/icons/hicolor/*/apps/freestyle.png`
- [x] Verified generated desktop entry has `StartupWMClass=freestyle`


matteo8p (approved):
This looks fantastic. Ship it.

PR #257, opened 2026-06-13
fix: move toLowerCase() outside loop in post-process and formats

### LINKED ISSUE #253: Unnecessary `toLowerCase()` calls in a loop (easy)
# Overview
We have a bad code smells, where `toLowerCase()` is being called inside of a loop, when it would be more efficient to call it one time outside of the loop. Very small change, but prevents new strings from having to be created in memory.

`apps/server/src/routes/formats.ts` (match endpoint, ~lines 82–88)

# How to fix
1. In `getContextHint` (post-process.ts), add `const matchStrLower = matchStr.toLowerCase();` before the loop and use it inside.
2. Apply the same change to the equivalent loop in `apps/server/src/routes/formats.ts` (the variable there is context).

Ensure that transcription is still working and the "Vocabulary" feature in Freestyle still works. 

# First time contributing? 

<p align="left">
</p>

kev0-4:
Fixes [#253](https://github.com/freestyle-voice/freestyle/issues/253)
  
  What changed

  Moved toLowerCase() calls outside of loops in two places to avoid creating unnecessary strings on every iteration.

  - apps/server/src/lib/post-process.ts (getContextHint): extracted matchStr.toLowerCase() into const matchStrLower before the loop
  - apps/server/src/routes/formats.ts (/match endpoint): extracted context.toLowerCase() into const contextLower before the loop

### DIFF
--- a/apps/server/src/lib/post-process.ts
+++ b/apps/server/src/lib/post-process.ts
@@ -40,6 +40,7 @@ function getContextHint(
 
   const matchStr = buildMatchContext(rawContext);
   if (!matchStr) return "";
+  const matchStrLower = matchStr.toLowerCase();
 
   try {
     const rows = db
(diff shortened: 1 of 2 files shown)


matteo8p:
@kev0-4 I dropped a comment [here](https://github.com/freestyle-voice/freestyle/issues/253#issuecomment-4699411797) about this PR! TLDR: I want to limit one good first issue per person. Hope you understand!

PR #259, opened 2026-06-13
fix: deduplicate regex compilation in dictionary replacement loop

### LINKED ISSUE #252: Remove duplicate regex code (Easy)
# Overview 

In the dictionary-replacement loop, the same regex is compiled twice per dictionary entry — once for `.test()` and again for `.replace()`. We want to just unify it into a single regex. 

See [here](https://github.com/freestyle-voice/freestyle/blob/main/apps/server/src/lib/post-process.ts#L215C1-L216C1) and [here](https://github.com/freestyle-voice/freestyle/blob/main/apps/server/src/lib/post-process.ts#L219)

# How to fix
1. Open `apps/server/src/lib/post-process.ts` and find the dictionary loop.
2. Replace the test-then-replace pattern with a single .replace() call that tracks whether anything matched:
```
const regex = new RegExp(`\\b${escaped}\\b`, "gi");
let matched = false;
cleanedText = cleanedText.replace(regex, () => {
  matched = true;
  return value;
});
if (matched) matchedIds.push(id);
```
Or something like that.

The acceptance criteria is that now we are only using a single regex and all test cases passed. Also make sure that the dictionary feature works. This change could affect the dictionary feature. 

# First time contributing? 

<p align="left">
</p>

kev0-4:
Fixes #252 

What changed

- In the dictionary replacement loop in apps/server/src/lib/post-process.ts, the same regex was being compiled twice per dictionary entry — once for .test() and again for .replace().

- Replaced the test-then-replace pattern with a single .replace() call using a callback to track whether a match occurred, so the regex is only compiled and run once per entry.

### DIFF
--- a/apps/electron/src/main/linux-terminal-focus.ts
+++ b/apps/electron/src/main/linux-terminal-focus.ts
@@ -1,0 +1,262 @@
+import { execFile } from "node:child_process";
+import { readFile } from "node:fs/promises";
+import { createAppLogger } from "@freestyle/utils";
+
+const log = createAppLogger("paste");
+
+// Keep in sync with terminal_classes[] in native/linux-fast-paste.c
+const TERMINAL_IDENTIFIERS = [
+  "konsole",
+  "gnome-terminal",
+  "terminal",
+  "kitty",
+  "alacritty",
+  "terminator",
+  "xterm",
+  "urxvt",
+  "rxvt",
+  "tilix",
+  "terminology",
+  "wezterm",
+  "foot",
+  "st-256color",
+  "st-",
+  "yakuake",
+  "ghostty",
+  "guake",
+  "tilda",
+  "hyper",
+  "tabby",
+  "sakura",
+  "warp",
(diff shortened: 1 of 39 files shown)


matteo8p (changes requested):
Hey Kevin, thank you so much for putting this in. It looks good. Looks like we have a merge conflict. Could you resolve that? I'm happy to merge in the PR after that's resolved!

kev0-4:
After resolving the merge conflict, I can see that the upstream refactor already moved the dictionary replacement logic to dictionary-replacements.ts and solved the double-regex issue using a string comparison approach

matteo8p:
@kev0-4 Ahhh shoot, that's my bad. I think my coding agent caught that and fixed it, and I didn't catch that. That's my bad. Do you want to try taking on this issue instead? 

https://github.com/freestyle-voice/freestyle/issues/253

matteo8p:
Closing this out since it's already fixed

PR #268, opened 2026-06-14
fix(linux): use persistent uinput helper for Wayland paste

### LINKED ISSUE #266: Paste on Hyprland is not working
## The Problem
Despite now a better linux paste system in place, it still fails on hyprland

## The Reason
`linux-fast-paste` creates a temporary uinput keyboard, waits only `50 ms`, emitts `Ctrl+V`, then destroys it almost immediately. Hyprland has not finished registering the keyboard, so events are silently dropped despite exit code 0.

xikhar:
## Summary

- This fixes #266 
- Add a persistent Linux Wayland `uinput` paste helper.
- Create the virtual keyboard once at startup and reuse it for paste commands.
- Support normal and terminal paste shortcuts.
- Fall back to `wtype` when `uinput` is unavailable or fails.
- Handle helper timeouts, crashes, serialized commands, and graceful shutdown.
- Preserve existing macOS, Windows, and Linux X11 behavior.

## Verification

- Verified repeated normal and terminal pastes on Hyprland.
- Verified one-shot `uinput` compatibility.
- Passed Biome checks, TypeScript checks, and production build.

### DIFF
--- a/apps/electron/native/linux-fast-paste.c
+++ b/apps/electron/native/linux-fast-paste.c
@@ -11,6 +11,7 @@
  * Usage:
  *   linux-fast-paste                 # XTest mode (X11)
  *   linux-fast-paste --uinput        # uinput mode (Wayland)
+ *   linux-fast-paste --uinput-server # persistent uinput mode over stdin/stdout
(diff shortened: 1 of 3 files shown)


matteo8p (approved):
Ship it

PR #274, opened 2026-06-15
Clearer Settings + Format Menus

tigerisaac:
Basically, this includes some changes for making our settings page a lot clearer. In the past, you just had to scroll down with a bunch of headers, which made for a very complicated, very unorganized settings page. This PR will add a sidebar in the settings page that can allow the user to click through all the subsections of it.
image

Additionally, it will introduce a new language picker to allow for multiple languages to be selected, and just a cleaner way than the dropdown we had before.
image

And the final change is in the formats page. It was really bugging me that when you have to edit a format, the editing UI pops up at the top of the formats page, rather than directly under the card of the format that you're trying to edit. So it's now updated so that the UI will show up under the card directly.
image

PR was written by Freestyle. Yay


tigerisaac:
Actually, I'll come back w/ an issue and make the PR better scoped.

PR #296, opened 2026-06-17
Fix Linux whisper-server AVX-512 crash

### LINKED ISSUE #289: Since version 0.2.0, the local Whisper feature no longer starts.
Image

### Steps to Reproduce / Tried Solutions
- Redownloaded the model -> Same error.
- Switched to the "small" model -> Same error.

### Environment
- **Installation Type:** AppImage
- **OS:** CachyOS (Kernel: `7.0.12-1-cachyos`)

andrewjsalzer:
Fixes #289

## Summary
- Build bundled Linux whisper.cpp binaries with `GGML_NATIVE=OFF` so GitHub Actions runner CPU features like AVX-512 are not baked into release artifacts.
- Add a Linux CI verification step that disassembles the bundled `whisper-server` and fails if AVX-512/ZMM instructions are present.
- Install `binutils` in the Linux packaging job for the verifier's `objdump` dependency.

## Root cause
Linux release `whisper-server` was built with `GGML_NATIVE=ON` on CI runners that expose AVX-512. The binary embeds AVX-512 instructions in `ggml_cpu_init()` that run unconditionally at startup. On common consumer CPUs without AVX-512 (e.g. Intel 12th/13th gen), the process crashes with SIGILL right after the GPU init log lines — matching the reports in #289.

Disabling GPU/flash attention does not fix it; this is a CPU instruction-set portability issue, not a GPU driver issue.

## Test plan
- Verified the local failure mode: bundled `whisper-server` crashes with SIGILL in `ggml_cpu_init()` on an AVX2-only CPU, including with `--no-gpu --no-flash-attn`.
- Verified helper output adds `-DGGML_NATIVE=OFF` only for Linux `--resources`/release builds.
- Built a local AppImage from this branch; `whisper-server` starts and reaches `Server ready on port 8178` on the same machine where 0.2.0 failed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,7 +135,7 @@ jobs:
         working-directory: apps/electron
         run: node scripts/download-whisper-cpp.mjs --resources
 
-      - name: Verify whisper-server is portable (no AVX-512)
+      - name: Verify whisper binaries are portable (no AVX-512)
         working-directory: apps/electron
         run: node scripts/verify-whisper-portable.mjs
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,7 +129,7 @@ jobs:
         run: pnpm --filter @freestyle/server build
 
       - name: Install native build dependencies
-        run: sudo apt-get update && sudo apt-get install -y libx11-dev libxtst-dev libxext-dev libglib2.0-dev cmake libx11-xcb-dev
+        run: sudo apt-get update && sudo apt-get install -y libx11-dev libxtst-dev libxext-dev libglib2.0-dev cmake libx11-xcb-dev binutils
 
       - name: Build whisper.cpp binaries
         working-directory: apps/electron
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,6 +135,10 @@ jobs:
         working-directory: apps/electron
         run: node scripts/download-whisper-cpp.mjs --resources
 
+      - name: Verify whisper-server is portable (no AVX-512)
+        working-directory: apps/electron
+        run: node scripts/verify-whisper-portable.mjs
+
       - name: Build electron (Linux)
         working-directory: apps/electron
         run: pnpm run build:linux
--- a/apps/electron/scripts/download-whisper-cpp.mjs
+++ b/apps/electron/scripts/download-whisper-cpp.mjs
@@ -99,7 +99,8 @@ async function buildFromSource(outDir) {
 
   console.log("Building (this may take a minute)...");
   mkdirSync(buildDir, { recursive: true });
-  execFileSync("cmake", getWhisperCmakeArgs(), {
+  const forBundledRelease = process.argv.includes("--resources");
+  execFileSync("cmake", getWhisperCmakeArgs({ forBundledRelease }), {
     cwd: buildDir,
     stdio: "inherit",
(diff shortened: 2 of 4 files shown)


andrewjsalzer:
Continuing to test this. Please hold.

andrewjsalzer:
This PR addresses #289.

**Root cause:** the Linux 0.2.0 `whisper-server` was built on CI with `-march=native` (AVX-512 on GitHub runners). On CPUs without AVX-512, it crashes with SIGILL during `ggml_cpu_init()` — the same failure shown in the issue screenshot, right after the GPU init log lines.

**Fix:** build bundled Linux whisper binaries with `-DGGML_NATIVE=OFF` and add a CI check that the shipped binary contains no AVX-512 instructions.

**Verified locally:** AppImage built from this branch starts whisper successfully on an i5-12500H (AVX2, no AVX-512), where the installed 0.2.0 build previously failed.

matteo8p:
Reviewing this later today.

MathurAditya724 (commented):
Nice fix — the root cause makes sense and the diagnosis is a lot more convincing than the earlier flash-attn/GPU theories. Building with `-march=native` on a runner that happens to have AVX-512 and then shipping that binary to everyone is exactly the kind of thing that bites only a subset of users, and the SIGILL-at-startup in `ggml_cpu_init()` lines up with the screenshot in the issue. I also like that you scoped it to Linux `--resources` so macOS (which also runs `--resources`) and local dev builds aren't affected, and that you added a CI guard instead of just patching it once.

A few things I'd tweak before merging:

1. The verifier only checks `whisper-server`, but we also bundle `whisper-cli` and it comes out of the same build. It's fixed by this change too, but the guard won't notice if the CLI ever regresses on its own. Probably worth just looping over both binaries.

2. `getWhisperCmakeArgs({ forBundledRelease })` takes that option but the caller never passes it — it falls back to reading `process.argv` inside the helper instead. Right now `forBundledRelease` is dead, and reaching into `process.argv` from inside the function makes it awkward to test. I'd either thread an explicit flag through from the caller or drop the param.

3. The `zmm|evex` regex over `objdump` output is a decent first line of defense, but it's a bit fragile — it can false-positive on a symbol name that happens to contain those substrings, and it can miss EVEX-encoded ops that only touch xmm/ymm or the mask registers (k0–k7) without naming a zmm. Not a blocker, but maybe also scan for `%k[0-7]` usage, or at least drop a comment noting it's a best-effort check.

One thing I double-checked and think you got right: the runtime source build in `apps/server/src/lib/whisper/models.ts` is left alone, which is correct since that one compiles on the user's own CPU and should keep native optimizations.

Overall this looks good to me — approving once the above are sorted (or with a note on why not). Thanks for tracking it down and testing on the actual affected hardware.

andrewjsalzer:
@MathurAditya724 Thanks for the thorough review — addressed all three items in e7b9a4f:

1. **Both binaries** — the verifier now loops over `whisper-server` and `whisper-cli`.
2. **Explicit flag** — `download-whisper-cpp.mjs` passes `{ forBundledRelease }` into `getWhisperCmakeArgs()`; the helper no longer reads `process.argv`.
3. **Broader pattern + comment** — the objdump scan also matches `%k[0-7]` mask registers, with a note that it's a best-effort guard against the GGML_NATIVE regression.

Ready for another look when you have a moment.

matteo8p (approved):
Ship it

PR #317, opened 2026-06-22
feat(sdk): plugin system

MathurAditya724:
Adds a Vite-inspired plugin system that lets users extend the dictation pipeline. Introduces the `@freestyle/sdk` contract package (types + a host-agnostic loader/registry) and wires plugins into **both hosts**: the server (transcription/cleanup pipeline) and the Electron main process (output delivery).

## What's included
- **`@freestyle/sdk`** — `Plugin` is a named object (`{ name, enforce?, apply?, setup?, ...hooks }`). Hooks: `afterTranscribe`, `beforeCleanup`, `afterCleanup`, `beforeOutput`, plus `event`/`config`/`dispose`. `OutputMode` const object (`paste`/`copy`/`none`), a discriminated `FreestyleEvent` union, `transform()` and `sortPlugins()` helpers, and a host-agnostic `PluginRegistry` + `loadPlugins()` core.
- **Loaders (server + Electron-main)** — both consume the shared SDK core, injecting host-specific bits (context builder, settings/dir source, logger, error reporter). They load plugins from npm specifiers + a local `<userData>/plugins/` dir, flatten presets, host-filter via `apply`, run `setup`, and order by `enforce`. Handlers are isolated so a bad plugin can't crash a dictation.
- **Server pipeline** — `afterTranscribe` + `transcribed` event in the transcribe route; `beforeCleanup` (prompt fragments + register override), `afterCleanup` (final rewrite chain), and the `cleaned` event in post-process.
- **Electron-main** — `beforeOutput` at the paste/copy IPC seam (rewrite text or switch delivery between paste/copy/none; `none` suppresses output), threading the renderer's captured app context through to the hook; emits `outputDelivered`. `recordingStarted/Committed/Cancelled` emitted from the hotkey/escape seams.
- **`plugins` setting schema** in `@freestyle/validations`.

## Testing
`pnpm test` (161 pass), `tsc --noEmit` (sdk + server + electron node/web), `biome check`, and builds (`pnpm build` server → dist, `electron-vite build` main/preload/renderer) all green.


MathurAditya724:
fix-ci: attempt 1 — Typecheck and server tests can't resolve `@freestyle/utils`/`@freestyle/sdk` because the new pkgroll `exports` point only at `dist/`, which isn't built before `pnpm --filter @freestyle/server build`/vitest run in CI. Restoring the `development`/`source` → `src/index.ts` conditions so the packages stay resolvable from source in-repo while keeping the dist build for publishing.

PR #322, opened 2026-06-23
Use Better Auth client for cloud sign-in

MathurAditya724:
Switch the desktop cloud sign-in device flow to Better Auth's client/device plugin instead of manually calling the device endpoints. Token ownership remains in Electron main so the renderer still only receives the user profile.

## Testing
- pnpm --filter @freestyle/server build
- pnpm --filter @freestyle/electron typecheck


MathurAditya724:
This is ready for review. I ran a full self-review pass, fixed the issues found in the Freestyle Cloud auth/model flows and onboarding, and verified with server tests/build plus Electron typecheck. The visible PR check is passing.

PR #323, opened 2026-06-23
feat(settings): add history pause control

MathurAditya724:
Adds a Data-tab switch to pause saving new transcription history and centralizes server history writes so batch and streaming paths honor the setting. Includes locale strings and server tests for paused/unpaused history persistence.

## Testing
- pnpm --filter @freestyle/electron typecheck:web
- pnpm --filter @freestyle/server build
- pnpm --filter @freestyle/server test
- pnpm exec biome check changed files
- git diff --check


MathurAditya724:
Self-review is clean and CI is green. I also resolved review-pass issues before opening: centralized history writes to avoid duplicated pause logic, normalized nullable SQLite fields, added paused/unpaused history persistence tests, and added the new Data-tab strings across supported locales.

MathurAditya724:
Removed the duplicate dev badge I added. I searched the checked-out source and fetched refs for an existing renderer DEV badge and could not find one in origin/main; I do not want this PR to add a second badge in the wrong place, so this PR now scopes to the history pause feature only.

PR #324, opened 2026-06-24
feat(plugins): plugin UI ecosystem + audio-transcription plugin

MathurAditya724:
Lets plugins contribute sandboxed UI pages, managed from a new **Plugins hub**. Keeps Freestyle a lean dictation core: features (even first-party ones) ship as installable plugins. First-party plugins live in a new root-level `plugins/` workspace and are distributed as `@freestyle/*` npm packages — the same path third parties will use, so this dogfoods the real plugin authoring/distribution workflow.

## What's included
- **SDK** — `PluginUIPage` manifest types + `parsePluginPages` (reads `freestyle.contributes.pages` from a plugin's `package.json`), and the `FreestyleBridge` (`window.freestyle`) contract: a pre-authed server `api()`, plus host `invoke()`/`on()` for actions (paste/copy/toast/navigate) and events.
- **Electron main** — a `freestyle-plugin://` protocol serving a plugin's built `ui/` (path-traversal-guarded), package-manifest discovery, a `plugins:list` IPC, and a `PluginViewManager` that overlays a sandboxed `WebContentsView` on the dashboard window. A dedicated `plugin-bridge` preload injects the bridge + server config + theme tokens (`contextIsolation`, no `nodeIntegration`).
- **Renderer** — a `/plugins` hub (Installed tab with enable/Open cards; Browse tab stubbed "coming soon") and a `<PluginPage>` host that bounds-syncs the native view to a placeholder div (resize/scroll-aware). New nav entry + i18n.
- **Flagship `@freestyle/plugin-audio-transcription`** — drop audio files to transcribe them via `/api/transcribe`, rendered as raw/cleaned text with copy, themed via injected design tokens. Seeded into the default `plugins` setting for new installs.

Webview model follows the proven VS Code/Figma blueprint. Runtime npm install-at-runtime (so Browse → Install works in the packaged app) is a deliberate follow-up; V1 fully supports local-dir and workspace-resolved plugins.

## Testing
`pnpm test` (175 pass), `tsc --noEmit` (sdk + server + electron node/web + plugin), `biome check`, and builds: server → dist, plugin → `dist` + `ui/dist`, `electron-vite build` (main/preload incl. `plugin-bridge`/renderer) — all green.


MathurAditya724:
Ready for review. All CI is green — Typecheck, Lint, Server API tests (175), Electron E2E, and Build on Linux/Windows/macOS + the server image.

Self-review surfaced and fixed a set of issues before promotion, notably:
- Scoped package names (`@freestyle/...`) broke both the page route and the `freestyle-plugin://` host — pages are now routed/served by a URL-safe `slug` (with SDK tests incl. a URL round-trip).
- The bridge config (incl. the server token) was passed via process args; it now flows over IPC so the token can't leak into process listings.
- UI plugin discovery now reads the `plugins` setting over HTTP, so it works with a remote server.
- Unresolved default-plugin specifiers warn instead of erroring on every startup; dropped the `paste` host action (wrong target for an in-window page) and the unimplemented event surface.

Follow-ups (out of scope, by design): runtime npm install in the packaged app so Browse → Install works end-to-end, plus a hosted registry.

PR #344, opened 2026-06-29
fix: stop whisper server before deleting model file

### LINKED ISSUE #205: deleted local models are not actually deleted - win 11 home 25H2
I noticed that when I'm switching to a different local model, the previous models which I tried stay in the system even if I click delete. I know that because the app shows how many models are in the system in the UI, and also, if I switch back to the previous model which I deleted, it doesn't download, app just switches to it instantly. That means it is stored locally and is not deleted. But if I click delete, I want the model to be actually deleted from my laptop, not to take space on my hard drive.

Freestyle version: 0.1.6
Laptop: VivoBook ASUS K3402ZA
Processor: 12th Gen Intel(R) Core(TM) i7-12700H (2.30 GHz)
Installed RAM: 16.0 GB (15.6 GB usable)
Graphics card: Intel(R) Iris(R) Xe Graphics (128 MB)
Storage: 685 GB of 954 GB used
System type: 64-bit operating system, x64-based processor
OS: Windows 11 Home 25H2
OS build: 26200.8524
Experience: Windows Feature Experience Pack 1000.26100.315.0

Image

Image

MathurAditya724:
On Windows, the whisper-server process holds the model file open, so `unlinkSync` fails with EPERM/EBUSY when the user clicks delete. The empty `catch {}` silently swallows the error, making the model appear deleted in the UI while it remains on disk.

This stops the server before unlinking, matching what `deleteMlxModel` already does. Uses a dynamic import to avoid a circular dependency between `models.ts` and `server.ts`.

## Testing

- `tsc --noEmit` — passes
- `vitest run` — 169 tests pass
- `biome check` — clean

Fixes #205

### DIFF
--- a/apps/server/src/lib/whisper/models.ts
+++ b/apps/server/src/lib/whisper/models.ts
@@ -252,11 +252,17 @@ export function cancelDownload(modelId: string): boolean {
   return true;
 }
 
-export function deleteModel(modelId: string): boolean {
+export async function deleteModel(modelId: string): Promise<boolean> {
   const model = getWhisperModel(modelId);
   if (!model) return false;
 
   cancelDownload(modelId);
+
+  // Stop the whisper server before deleting — on Windows the server
+  // process holds the model file open, so unlinkSync would fail with
+  // EPERM/EBUSY while it's running.
+  const { stopServer } = await import("./server.js");
+  await stopServer();
 
   const path = getModelPath(model);
   try {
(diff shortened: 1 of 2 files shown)


MathurAditya724:
all 10 CI checks green (Build macOS/Linux/Windows, Lint, Typecheck, Test Server API, Test Electron E2E, Build server image, Preview Changelog, CI Status). self-review looks clean — minimal async change to stop the whisper server before deleting model files. marked ready for review.

PR #357, opened 2026-06-30
feat: improve onboarding local model download UX

siddharthkul:
## Problem

First-time local users could reach the onboarding tutorial, or finish onboarding, before the on-device voice model finished downloading — or without knowing a download was needed at all. Silent setup surfaced only as a tiny status line, and failures like **\"Whisper model not downloaded\"** could appear as a native error dialog with no in-app recovery path.

Screenshot 2026-06-30 at 5 57 57 PM

Screenshot 2026-06-30 at 5 58 04 PM

### Local model setup panel
- Adds a `ModelSetupPanel` card on the **language** and **tutorial** onboarding steps for local-only users.
- **Idle:** explains the one-time local model download with a **Download (size)** button.
- **Downloading / building engine:** shows progress with bytes, speed, percent, and an indeterminate pulse state matching `voice-row.tsx`.
- **Error:** shows a human-readable error and **Try again** recovery path.
- Uses `aria-live="polite"` while setup is active.

### Onboarding flow
- Local-only users now start the model download explicitly from the setup panel.
- Tutorial **Finish** is disabled for local users until the selected local model is `ready`.
- Cloud users keep the previous frictionless behavior: Freestyle Cloud stays the default, the offline fallback downloads silently in the background, and the setup panel is hidden.
- Tracks the last explicit local model pick so fallback downloads target the model the user chose, even if they later switch back to cloud.
- Adds retry analytics with `onboarding_model_download_started` and `{ retry: true }`.

### i18n
- Adds `onboarding.modelSetup.*` keys in `en.json` and `template.json`.
- Removes unused `onboarding.tutorial.almostReady` from `en.json`.

## Testing

- [x] `pnpm --filter @freestyle-voice/electron typecheck:web`
- [x] `pnpm biome check apps/electron/src/renderer/src/onboarding.tsx apps/electron/src/renderer/src/components/model-setup-panel.tsx apps/electron/src/renderer/src/locales/en.json apps/electron/src/renderer/src/locales/template.json`
- [ ] `pnpm dev` → tray → **Reset Onboarding**
- [ ] Delete `~/.cache/freestyle/whisper-models/ggml-small-q5_1.bin` to simulate first run
- [ ] Local path: Language step shows the download card with **Download** button
- [ ] Local path: Tap Download → progress bar updates with bytes/percent
- [ ] Local path: Tutorial **Finish** stays disabled until download completes
- [ ] Cloud path: Sign in with Freestyle Cloud → no setup panel; offline fallback downloads silently
- [ ] Induce a download error → **Try again** works

## Follow-up

- Translate `onboarding.modelSetup.*` into non-English locale files.
- Remove stale `almostReady` keys from other locale files.

### DIFF
--- a/apps/electron/src/renderer/src/components/model-setup-panel.tsx
+++ b/apps/electron/src/renderer/src/components/model-setup-panel.tsx
@@ -1,0 +1,160 @@
+import type { VoiceItem } from "@renderer/lib/models";
+import { formatBytes, formatSpeed } from "@renderer/lib/models";
+import { cn, ON_DEVICE_PHRASE } from "@renderer/lib/utils";
+import { Download, HardDrive, Loader2, RefreshCw } from "lucide-react";
+import { useTranslation } from "react-i18next";
+import { Button } from "./ui/button";
+import { Progress } from "./ui/progress";
+
+type ModelSetupPanelProps = {
+  model: VoiceItem | undefined;
+  onDownload: () => void;
+  onRetry: () => void;
+  /** When true, offline setup is encouraged but not required to continue. */
+  optional?: boolean;
+};
+
+export function ModelSetupPanel({
+  model,
+  onDownload,
+  onRetry,
+  optional = false,
+}: ModelSetupPanelProps): React.JSX.Element | null {
+  const { t } = useTranslation();
+
+  if (!model || model.kind !== "local") return null;
+
+  const status = model.status ?? "not_downloaded";
+  if (status === "ready") return null;
+
+  const sizeLabel =
+    model.sizeBytes != null ? formatBytes(model.sizeBytes) : null;
+  const progress = model.state?.downloadProgress;
+  const hasProgress = !!progress;
+  const isActive =
+    status === "downloading" ||
+    status === "verifying" ||
+    model.state?.phase === "building_binary";
+  const isError = status === "error";
+  const isIdle = status === "not_downloaded" && !isActive;
+
+  return (
+    <div
+      className={cn(
+        "bg-card border-border mt-6 w-full rounded-[12px] border px-4 py-3.5",
+        isError && "border-destructive/40",
+      )}
+      role={isActive ? "status" : undefined}
+      aria-live={isActive ? "polite" : undefined}
+    >
+      <div className="flex items-start gap-3">
+        <div
+          className={cn(
+            "mt-0.5 flex h-8 w-8 shrink-0 items-center justify-center rounded-[8px]",
+            isError ? "bg-destructive/10" : "bg-accent",
+          )}
+        >
+          {isActive ? (
+            <Loader2 className="text-primary h-4 w-4 animate-spin" />
+          ) : isError ? (
+            <HardDrive className="text-destructive h-4 w-4" />
+          ) : (
+            <HardDrive className="text-primary h-4 w-4" />
+          )}
+        </div>
+
+        <div className="min-w-0 flex-1">
+          <p className="text-foreground text-[13.5px] leading-snug font-medium">
+            {isError
+              ? t("onboarding.modelSetup.errorTitle")
+              : isActive
+                ? model.state?.phase === "building_binary"
+                  ? t("onboarding.modelSetup.buildingEngine")
+                  : t("onboarding.modelSetup.downloading", {
+                      name: model.name,
+                    })
+                : optional
+                  ? t("onboarding.modelSetup.optionalTitle")
+                  : t("onboarding.modelSetup.title")}
+          </p>
+
+          <p className="text-muted-foreground mt-1 text-[12px] leading-snug">
+            {isError
+              ? (model.state?.error ?? t("onboarding.modelSetup.errorFallback"))
+              : isActive
+                ? model.state?.phase === "building_binary"
+                  ? t("onboarding.modelSetup.buildingEngineDesc")
+                  : t("onboarding.modelSetup.downloadingDesc", {
+                      phrase: ON_DEVICE_PHRASE,
+                    })
+                : t("onboarding.modelSetup.idleDesc", {
+                    name: model.name,
+                    size: sizeLabel ? ` (${sizeLabel})` : "",
+                    phrase: ON_DEVICE_PHRASE,
+                  })}
+          </p>
+
+          {isActive && (
+            <div className="mt-3 space-y-1.5">
+              <Progress
+                value={hasProgress ? (progress?.percent ?? 0) : undefined}
+                className={cn("h-[6px]", !hasProgress && "animate-pulse")}
+              />
+              <div className="text-muted-foreground mono flex justify-between text-[10px]">
+                {model.state?.phase === "building_binary" && !hasProgress ? (
+                  <span>{t("onboarding.modelSetup.buildingEngine")}</span>
+                ) : hasProgress ? (
+                  <>
+                    <span>
+                      {formatBytes(progress!.bytesDownloaded)} /{" "}
+                      {formatBytes(progress!.bytesTotal)}
+                    </span>
+                    <span>
+                      {progress!.speedBps > 0 &&
+                        formatSpeed(progress!.speedBps)}
+                      {progress!.percent > 0 && ` · ${progress!.percent}%`}
+                    </span>
+                  </>
+                ) : (
+                  <span>{t("onboarding.modelSetup.preparing")}</span>
+                )}
+              </div>
+            </div>
+          )}
+
+          {isIdle && (
+            <Button
+              variant="default"
+              size="sm"
+              className="mt-3"
+              onClick={onDownload}
+            >
+              <Download data-icon="inline-start" />
+              {sizeLabel
+                ? t("onboarding.modelSetup.downloadButton", { size: sizeLabel })
+                : t("onboarding.modelSetup.downloadButtonShort")}
+            </Button>
+          )}
+
+          {isError && (
+            <Button
+              variant="outline"
+              size="sm"
+              className="mt-3"
+              onClick={onRetry}
+            >
+              <RefreshCw data-icon="inline-start" />
+              {t("onboarding.modelSetup.retry")}
+            </Button>
+          )}
+
+          {optional && isIdle && (
+            <p className="text-muted-foreground mono mt-2 text-[10px] tracking-wide uppercase">
+              {t("onboarding.modelSetup.optionalHint")}
+            </p>
+          )}
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/apps/electron/src/renderer/src/components/model-setup-panel.tsx
+++ b/apps/electron/src/renderer/src/components/model-setup-panel.tsx
@@ -10,15 +10,12 @@ type ModelSetupPanelProps = {
   model: VoiceItem | undefined;
   onDownload: () => void;
   onRetry: () => void;
-  /** When true, offline setup is encouraged but not required to continue. */
-  optional?: boolean;
 };
 
 export function ModelSetupPanel({
   model,
   onDownload,
   onRetry,
-  optional = false,
 }: ModelSetupPanelProps): React.JSX.Element | null {
   const { t } = useTranslation();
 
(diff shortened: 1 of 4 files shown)


apps/electron/src/renderer/src/onboarding.tsx
```diff
@@ -677,32 +670,27 @@ export default function OnboardingPage(): React.JSX.Element {
...
-      chosen.status === "verifying" ||
-      chosen.status === "not_downloaded"
-    ) {
-      const p = chosen.state?.downloadProgress;
-      const pct = p?.bytesTotal ? ` ${p.percent}%` : "";
-      return `Downloading ${chosen.name}${size}, your private transcription model…${pct}`;
-    }
-    return null;
-  })();
+  const localSetupModel = chosen?.kind === "local" ? chosen : recommended;
```
MathurAditya724: **Bug (low-med):** When the user selects a cloud model, `chosen.kind === "cloud"` so `localSetupModel` falls back to `recommended`. This means if the user previously opened the selector, picked a *different* local model, then switched to cloud — the panel still shows the *recommended* model (MLX Qwen / Whisper small), not the one they explicitly chose. `startLocalDownload` and `retryLocalDownload` would then trigger a download for the wrong model.

In practice few users will follow this path during first-run, but a `lastSelectedLocalModel` ref tracking the most recent explicit local pick would be more correct.

apps/electron/src/renderer/src/onboarding.tsx
```diff
@@ -677,32 +670,27 @@ export default function OnboardingPage(): React.JSX.Element {
...
+    localSetupModel?.kind === "local" &&
+    (localSetupModel.status === "downloading" ||
+      localSetupModel.status === "verifying" ||
+      localSetupModel.state?.phase === "building_binary");
+
+  const startLocalDownload = useCallback(() => {
+    if (!localSetupModel?.defId || window.api?.isE2E) return;
+    capture("onboarding_model_download_started", {
+      model_id: localSetupModel.modelId,
+    });
```
MathurAditya724, line 719: **Nit (analytics gap):** `retryLocalDownload` does not fire any analytics event, while `startLocalDownload` fires `onboarding_model_download_started`. A retry after a failure is arguably *more* interesting to track. Consider firing `onboarding_model_download_retried` or at least reusing `onboarding_model_download_started` with `{ retry: true }`.

apps/electron/src/renderer/src/onboarding.tsx
```diff
@@ -677,32 +670,27 @@ export default function OnboardingPage(): React.JSX.Element {
...
+  const localSetupOptional = !!cloudUser && chosen?.kind === "cloud";
+  const mustHaveLocalReady = chosen?.kind === "local" && !chosenReady;
+  const localSetupActive =
+    localSetupModel?.kind === "local" &&
+    (localSetupModel.status === "downloading" ||
+      localSetupModel.status === "verifying" ||
+      localSetupModel.state?.phase === "building_binary");
+
+  const startLocalDownload = useCallback(() => {
+    if (!localSetupModel?.defId || window.api?.isE2E) return;
```
MathurAditya724, line 716: **Nit (consistency):** `startLocalDownload` guards against `window.api?.isE2E` but `retryLocalDownload` does not. In practice E2E likely never hits the error→retry path, but worth aligning for consistency.

apps/electron/src/renderer/src/components/model-setup-panel.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  })}
+          </p>
+
+          {isActive && (
+            <div className="mt-3 space-y-1.5">
+              <Progress
+                value={hasProgress ? (progress?.percent ?? 0) : undefined}
+                className={cn("h-[6px]", !hasProgress && "animate-pulse")}
+              />
+              <div className="text-muted-foreground mono flex justify-between text-[10px]">
```
MathurAditya724, line 98: **Nit:** When there is no progress data, the `Progress` bar gets `value={undefined}`. In `voice-row.tsx` the equivalent indeterminate state passes `value={100}` so the *full bar* pulses, which gives better visual feedback than an empty pulsing track. Consider aligning:
```tsx
<Progress
  value={hasProgress ? (progress?.percent ?? 0) : 100}
  className={cn("h-[6px]", !hasProgress && "animate-pulse")}
/>
```

apps/electron/src/renderer/src/onboarding.tsx
```diff
@@ -769,8 +757,10 @@ export default function OnboardingPage(): React.JSX.Element {
           <LanguageStep
             language={language}
             onSelect={saveLanguage}
-            setupStatus={setupStatus}
-            setupError={setupError}
+            localModel={localSetupModel}
```
MathurAditya724: **Design concern (cloud UX):** For Freestyle Cloud users, this renders a prominent `ModelSetupPanel` card with a download button on the language step *and* the tutorial step — two consecutive screens showing a card for something that is entirely optional.

The old flow was arguably better for cloud users: it silently downloaded the offline fallback in the background and the user never had to think about it. The new card adds visual noise and a decision point ("should I download this?") to what should be frictionless onboarding.

Some options:
1. **Hide `ModelSetupPanel` entirely for cloud users** — don't render the panel when `chosen.kind === "cloud"`. They can set up offline later from settings.
2. **Show the card only on the tutorial step**, not both steps.
3. **Keep the auto-download for cloud users, use the new panel only for local users** — cloud users get the silent background download (existing behavior), local-only users get the explicit panel (new behavior). Best of both worlds.

MathurAditya724 (commented):
Thanks for the PR — this is a meaningful UX improvement. The explicit download panel with progress, error recovery, and gated finish is much better than the silent background download with native error dialogs.

A few issues below, the biggest being the cloud-user experience: the `ModelSetupPanel` card now appears on **two consecutive onboarding steps** for something that's entirely optional for cloud users. The old flow silently downloaded in the background and cloud users never had to think about it. Consider either hiding the panel entirely for cloud users (`chosen.kind === "cloud"`) or keeping the auto-download for cloud users while using the new panel only for local-only users.

apps/electron/src/renderer/src/onboarding.tsx
siddharthkul: Good catch — added `lastLocalSetupRef` to track the most recent explicit local pick, so if someone picks a different on-device model in the selector and then switches to cloud, the fallback download targets the right model. Addressed in 36372dc.

apps/electron/src/renderer/src/onboarding.tsx
siddharthkul: Agreed — `retryLocalDownload` now fires `onboarding_model_download_started` with `{ retry: true }` so retries stay in the same funnel. Addressed in 36372dc.

apps/electron/src/renderer/src/onboarding.tsx
siddharthkul: Aligned — `retryLocalDownload` now has the same `window.api?.isE2E` guard as `startLocalDownload`. Addressed in 78e340e.

apps/electron/src/renderer/src/components/model-setup-panel.tsx
siddharthkul: Done — indeterminate state now uses `value={100}` with pulse, matching `voice-row.tsx`. Addressed in 36372dc.

apps/electron/src/renderer/src/onboarding.tsx
siddharthkul: Went with option 3: cloud users get the silent background auto-download again and the setup panel is hidden for them. Local-only users keep the explicit panel with progress, retry, and Finish gating. Addressed in 36372dc.

MathurAditya724: looks good — cloud users getting the silent download back while local-only keeps the explicit panel is the right call. the `lastLocalSetupRef` tracking is a clean way to handle the model-switch-then-cloud scenario too. thanks for the thorough fix.

apps/electron/src/renderer/src/components/model-setup-panel.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const isIdle = status === "not_downloaded" && !isActive;
+
+  return (
+    <div
+      className={cn(
+        "bg-card border-border mt-6 w-full rounded-[12px] border px-4 py-3.5",
+        isError && "border-destructive/40",
+      )}
+      role={isActive ? "status" : undefined}
+      aria-live={isActive ? "polite" : undefined}
```
MathurAditya724, line 45: **Nit:** Our card/panel radius is `rounded-[14px]` per DESIGN.md (§4). This uses `rounded-[12px]` — mind bumping it to 14 for consistency?

apps/electron/src/renderer/src/components/model-setup-panel.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { useTranslation } from "react-i18next";
+import { Button } from "./ui/button";
+import { Progress } from "./ui/progress";
+
+type ModelSetupPanelProps = {
+  model: VoiceItem | undefined;
+  onDownload: () => void;
+  onRetry: () => void;
+  /** When true, offline setup is encouraged but not required to continue. */
+  optional?: boolean;
```
MathurAditya724: **Nit (dead code):** Now that cloud users never see this panel, the `optional` prop and the `optionalTitle` / `optionalHint` i18n keys are unused. Totally fine to leave them if you want to keep the door open for later, but if not — removing them would keep things tidy.

MathurAditya724 (commented):
Nice work on the follow-up commits — all the earlier feedback is addressed cleanly. Two small things left and then this is good to merge.

MathurAditya724 (commented):
clean PR — typecheck and biome both pass, the cloud/local split is well-handled, and the finish-gating logic is solid.

a few things i traced through that all checked out:

- `localSetupModel` gracefully falls back through chosen → lastLocalSetupRef → recommended, and the panel early-returns on `undefined` / `kind !== "local"` / `status === "ready"`
- cloud users never see the panel (`showLocalSetupPanel = !cloudUser`) and still get the silent background download
- `canFinish` correctly unblocks if a local user switches to a cloud model via the selector
- the indeterminate progress bar (`value={100}` + `animate-pulse`) matches the existing pattern in `voice-row.tsx`

the `optional` prop on `ModelSetupPanel` isn't used by either call site right now — fine if it's intentional scaffolding for a future path, just noting it.

stale `almostReady` keys in non-English locales are called out in the follow-up section, so no concern there.

apps/electron/src/renderer/src/components/model-setup-panel.tsx
siddharthkul: Done — bumped the setup panel container to `rounded-[14px]` to match DESIGN.md. Addressed in b3d78bf.

apps/electron/src/renderer/src/components/model-setup-panel.tsx
siddharthkul: Agreed — since cloud users no longer render this panel, I removed the unused `optional` prop plus `optionalTitle` / `optionalHint` from the locale files. Addressed in b3d78bf.

MathurAditya724 (commented):
code looks solid — went through the diff, types, and state flow carefully.

a few notes from the read:

- the `localSetupModel` fallback chain (chosen local → last explicit pick → recommended) handles the cloud-user-who-briefly-picked-local edge case well. `lastLocalSetupRef` is a nice touch.
- `else if (engine === "whisper")` change in `selectLocalModel` is correct — the old bare `else` would have set whisper state when `engine` was `undefined`.
- non-null assertions on `progress!` fields are safe since they're gated behind `hasProgress` (`!!progress`).
- `aria-live="polite"` + `role="status"` only when active is the right call for screen readers.
- stale `almostReady` keys in fr/pt/it/ja/de/es are fine as a follow-up (noted in the PR description).

one minor observation: `startLocalDownload` and `retryLocalDownload` are identical except for `{ retry: true }` in the analytics payload. could be a single callback with a `retry` param, but it's two small functions and the intent reads clearly as-is — not blocking on it.

PR #358, opened 2026-06-30
fix: prevent Globe/Fn hotkey during modifier chords

siddharthkul:
## Summary
- Globe/Fn dictation starts only after a short solo-key grace window, so modifier chords like Command+Globe or Option+Globe do not start recording.
- The macOS native listener emits `FN_DOWN` for solo Globe and `FN_DOWN:mods` when modifiers are present, preserving compound hotkeys and hotkey rebinding.
- Focused tests cover solo Fn activation, modifier-first chords, Fn-first near-simultaneous chords, post-activation modifiers, and recorder parsing.

## Test plan
- [x] Solo Globe press starts dictation
- [x] Command + Globe does not start dictation
- [x] Option + Globe does not start dictation
- [x] Globe held, then modifier added after activation continues recording until Globe release
- [x] Hotkey rebinding still captures compound shortcuts like Command+Fn
- [x] `corepack pnpm exec playwright test tests/key-listener.test.ts`
- [x] `corepack pnpm run typecheck:node`
- [x] `corepack pnpm exec biome check apps/electron/src/main/key-listener.ts apps/electron/src/main/hotkey-recorder.ts apps/electron/src/main/native-binary.ts apps/electron/tests/key-listener.test.ts`

### DIFF
--- a/apps/electron/native/macos-key-listener.swift
+++ b/apps/electron/native/macos-key-listener.swift
@@ -330,7 +330,17 @@ guard let monitor = NSEvent.addGlobalMonitorForEvents(matching: .flagsChanged, h
 
     if containsFn && !fnIsDown {
         fnIsDown = true
-        emit("FN_DOWN")
+        let otherMods = flags.intersection(modifierMask)
+        if otherMods.isEmpty {
+            emit("FN_DOWN")
+        } else {
+            var parts: [String] = []
+            if otherMods.contains(.control) { parts.append("control") }
+            if otherMods.contains(.option) { parts.append("option") }
+            if otherMods.contains(.shift) { parts.append("shift") }
+            if otherMods.contains(.command) { parts.append("command") }
+            emit("FN_DOWN:" + parts.joined(separator: ","))
+        }
     } else if !containsFn && fnIsDown {
         fnIsDown = false
         emit("FN_UP")
--- a/apps/electron/src/main/hotkey-recorder.ts
+++ b/apps/electron/src/main/hotkey-recorder.ts
@@ -210,8 +210,16 @@ export class HotkeyRecorder {
       return;
     }
 
-    if (line === "FN_DOWN") {
-      this.sendModifiers([...this.pendingModifiers, "Fn"]);
+    if (line === "FN_DOWN" || line.startsWith("FN_DOWN:")) {
+      const chordMods =
+        line === "FN_DOWN"
+          ? []
+          : line
+              .slice("FN_DOWN:".length)
+              .split(",")
+              .map((part) => MAC_FLAG_MODIFIERS[part.trim().toLowerCase()])
+              .filter((part): part is string => Boolean(part));
+      this.sendModifiers([...this.pendingModifiers, ...chordMods, "Fn"]);
       return;
     }
 
--- a/apps/electron/src/main/hotkey-recorder.ts
+++ b/apps/electron/src/main/hotkey-recorder.ts
@@ -217,6 +217,7 @@ export class HotkeyRecorder {
           : line
               .slice("FN_DOWN:".length)
               .split(",")
+              // Token set is defined by macos-key-listener.swift's FN_DOWN:mods emitter.
               .map((part) => MAC_FLAG_MODIFIERS[part.trim().toLowerCase()])
               .filter((part): part is string => Boolean(part));
       this.sendModifiers([...this.pendingModifiers, ...chordMods, "Fn"]);
--- a/apps/electron/src/main/key-listener.ts
+++ b/apps/electron/src/main/key-listener.ts
@@ -264,11 +264,14 @@ export class NativeKeyListener {
 
     const hotkey = this.options.hotkey.toLowerCase();
 
-    // Fn/Globe key
-    if (line === "FN_DOWN") {
+    // Fn/Globe key — FN_DOWN is solo only; FN_DOWN:mods tracks Fn held with chords.
+    if (line === "FN_DOWN" || line.startsWith("FN_DOWN:")) {
       this.macFnDown = true;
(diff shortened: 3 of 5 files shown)


MathurAditya724 (commented):
Reviewed the full event flow across the Swift listener, `key-listener.ts`, and `hotkey-recorder.ts`. Core approach is sound and consistent with existing patterns (`FN_DOWN:mods` mirrors the `RIGHT_MOD_DOWN:name` convention). The two unchecked test-plan items actually do hold up under trace:

- **Globe held, then modifier added** continues recording — the added modifier comes via the `FLAGS:` path; `checkMacCompoundRelease()` early-returns for fn/globe and `checkMacHotkeyMatch()` early-returns because the hotkey *is* fn/globe.
- **Rebinding Command+Fn** works — the recorder appends `chordMods` before `Fn`, so the captured combo includes the modifiers.

A few things to address:

**1. Edge-triggered emission only suppresses modifier-first ordering.**
`macos-key-listener.swift:205` emits the FN event only on Fn's rising edge (`containsFn && !fnIsDown`). So this reliably blocks "Command held → press Globe", but for "press Globe → then Command" (or near-simultaneous presses where Fn's `flagsChanged` lands first), a solo `FN_DOWN` is emitted and `onKeyDown()` fires *before* Command arrives — dictation starts anyway. Physical keypresses frequently land Fn-first. The first two checked test items only validate modifier-first ordering. Worth documenting this as a known limitation, or considering a short cancel-on-modifier-within-Nms window. As-is, the title ("solo press") overpromises relative to what's actually guaranteed.

**2. Unknown modifier tokens are silently dropped.**
In `hotkey-recorder.ts`, the `.map(... MAC_FLAG_MODIFIERS[...]).filter(Boolean)` silently discards any token not in the map. Safe today since Swift only emits control/option/shift/command, but please add a one-line comment noting the token set is contractually tied to the Swift emitter.

**3. No test coverage on any of the three changed paths.**
`apps/electron/tests/app.test.ts` doesn't touch FN handling. Given the subtle ordering/state logic, a unit test feeding synthetic lines (`FN_DOWN`, `FN_DOWN:command`, `FLAGS:command`, `FN_UP`) into `NativeKeyListener` / `HotkeyRecorder` and asserting `onKeyDown` / `onModifiers` would protect against regressions.

**Verified non-issues:** `modifierMask` correctly excludes `.function` (line 34), so `otherMods.isEmpty` is the right solo check; modifier ordering is consistent across Swift and TS (control, option, shift, command); both paths gate on `process.platform === "darwin"` before FN handling.

siddharthkul:
Addressed the review comments in `36afd3b`:

- Added a 50ms solo Fn/Globe grace window so Fn-first near-simultaneous chords can cancel before dictation starts.
- Gated `FN_UP` so ignored Fn chords don't emit a stray release.
- Added the requested recorder comment tying `FN_DOWN:mods` tokens to the Swift emitter contract.
- Added focused Playwright tests for solo Fn, modifier-first chords, Fn-first grace-window chords, post-activation modifiers, and recorder `FN_DOWN:mods` parsing.

Checks run:
- `corepack pnpm exec playwright test tests/key-listener.test.ts`
- `corepack pnpm run typecheck:node`
- `corepack pnpm exec biome check apps/electron/src/main/key-listener.ts apps/electron/src/main/hotkey-recorder.ts apps/electron/src/main/native-binary.ts apps/electron/tests/key-listener.test.ts`

MathurAditya724 (approved):
all three items addressed cleanly — grace window, emitter comment, and test coverage. the timer approach with re-validation in the callback is solid; both modifier arrival paths (`FLAGS:` and `RIGHT_MOD_DOWN:`) cancel correctly, and `FN_UP` gating prevents stray releases.

one optional suggestion: a "rapid Fn toggle" test (`FN_DOWN` → `FN_UP` within 50ms → assert `[]`) would cover the most common real-world scenario where the fix matters (quick Globe tap during a chord). not blocking though.

unrelated pre-existing nit: `KEY_UP:` is 7 chars but `line.slice(9)` is used at line 367 — should be `slice(7)`. not introduced here, just noticed while reading.

MathurAditya724 (commented):
solid approach — the three-layer defense (Swift-side `FN_DOWN:mods`, timer cancellation via `cancelMacSoloFnDownIfChorded`, and the `hasMacNonFnModifierActive()` guard in the timer callback) makes this robust against timing edge cases.

a few notes:

**pre-existing bug (not from this PR):** `KEY_UP:` handler at line 366 uses `line.slice(9)` but `"KEY_UP:"` is only 7 chars — should be `slice(7)`. `KEY_DOWN:` at line 362 correctly uses `slice(9)` since `"KEY_DOWN:"` is 9 chars. this means key-up events are silently broken for key names ≤2 chars (e.g. `KEY_UP:Up` → `""`) and wrong for longer ones (`KEY_UP:Space` → `"ace"`). might be worth a follow-up fix.

**test plan:** the two unchecked items in the PR description — "Globe held, then modifier added while holding" is actually covered by the automated test at line 75 (`adding a modifier after solo Fn activation keeps hold-to-talk active`). might want to check that box or note it.

**`app?.isPackaged`:** makes sense as a supporting change for the test file — without it, importing `NativeKeyListener` in a non-Electron context would crash on module load.

siddharthkul:
Added the optional rapid Fn toggle coverage in `03c007a`:

- `FN_DOWN` → `FN_UP` within the grace window now asserts no activation.
- Re-ran `corepack pnpm exec playwright test tests/key-listener.test.ts` and `corepack pnpm exec biome check apps/electron/tests/key-listener.test.ts`.

Agree the `KEY_UP:` slice nit looks pre-existing and unrelated; I’d keep that for a separate cleanup so this PR stays focused on Globe/Fn chord behavior.

PR #359, opened 2026-06-30
chore: add Knip CI check

siddharthkul:
## Summary
- Add Knip as a root dev dependency with a `pnpm run knip` script.
- Configure Knip for the pnpm monorepo entry points across Electron, server, packages, and plugins.
- Add a dedicated Knip job to Build & Test and include it in the aggregate CI Status gate.

## What is Knip?
Knip is a static analysis tool for finding unused files, unused dependencies, unlisted dependencies, unresolved imports, and unused exports in TypeScript/JavaScript projects. In this PR, it starts as a CI hygiene check focused on unused files/dependencies and import correctness for the monorepo.

## Review notes
- PR comments are intentionally not added; the standalone `Knip` check should be enough for users to inspect failures in CI.
- Export analysis is disabled initially because the repo has a larger existing export cleanup backlog; this PR starts with files/dependencies/import correctness.

## Test plan
- `corepack pnpm run knip`
- `corepack pnpm biome check package.json knip.jsonc .github/workflows/build.yml pnpm-lock.yaml`
- `corepack pnpm install --frozen-lockfile`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,24 @@ jobs:
       - name: Biome check
         run: pnpm biome check .
 
+  knip:
+    name: Knip
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: pnpm/action-setup@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: "package.json"
+          cache: "pnpm"
+
+      - run: pnpm install --frozen-lockfile
+
+      - name: Check unused files and dependencies
+        run: pnpm run knip
+
   typecheck:
     name: Typecheck
     runs-on: ubuntu-latest
@@ -298,6 +316,7 @@ jobs:
     if: always()
     needs:
       - lint
+      - knip
       - typecheck
       - test-server
       - test-electron
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: "22"
+          node-version-file: "package.json"
           cache: "pnpm"
 
       - run: pnpm install --frozen-lockfile
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version-file: "package.json"
+          node-version: "22"
           cache: "pnpm"
 
       - run: pnpm install --frozen-lockfile
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -59,7 +59,6 @@
       "ignoreDependencies": [
         "@hono/node-server",
         "@types/ws",
-        "node:sqlite",
         "playwright",
         "react-hotkeys-hook",
         "ws"
@@ -71,7 +70,12 @@
         "scripts/benchmark-post-process.ts",
         "tests/**/*.test.ts"
       ],
-      "project": ["src/**/*.{ts,tsx}", "scripts/**/*.ts", "tests/**/*.ts"],
+      "project": [
+        "src/**/*.{ts,tsx}",
+        "scripts/**/*.ts",
+        "tests/**/*.ts",
+        "*.config.ts"
+      ],
       "paths": {
         "freestyle-voice": ["../../packages/sdk/src/index.ts"],
         "@freestyle-voice/utils": ["../../packages/utils/src/index.ts"],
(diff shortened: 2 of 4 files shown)


MathurAditya724 (commented):
Solid addition. The CI wiring mirrors the existing `lint`/`typecheck` jobs cleanly and the `ci-status` gate is updated correctly. I cloned the branch, ran `pnpm install --frozen-lockfile` + `pnpm run knip` (node v22.20.0) — it passes, and a probe with a deliberately unused file/dep correctly fails the job, so the gate works. A few things worth tightening before merge:

**1. `node:sqlite` doesn't belong in `ignoreDependencies` (knip itself flags this).**
The run emits two configuration hints:
```
node:sqlite  apps/electron  knip.jsonc  Remove from ignoreDependencies
node:sqlite  apps/server    knip.jsonc  Remove from ignoreDependencies
```
`node:sqlite` is a Node builtin, so knip resolves it natively and the suppression is dead config. Drop it from both `ignoreDependencies` arrays. (`@types/ws`, `ws`, `playwright`, `react-hotkeys-hook`, `@hono/node-server`, `@types/json-schema` all check out as real, intentionally-suppressed deps — leave those.)

**2. Inconsistent `tests/**` handling between workspaces leaves `apps/server/tests/setup.ts` exposed.**
`apps/electron` lists `tests/**/*.ts` in *both* `entry` and `project`, so every test file is an entry. `apps/server` only makes `tests/**/*.test.ts` an entry while keeping `tests/**/*.ts` in `project`. `tests/setup.ts` is not a `.test.ts`, and the only thing referencing it is `apps/server/vitest.config.ts` via `setupFiles: ["tests/setup.ts"]` — which knip never analyzes, because the server `project` globs are `["src/**/*.{ts,tsx}", "scripts/**/*.ts", "tests/**/*.ts"]` and omit `*.config.ts` (electron includes `"*.config.ts"`; server doesn't).

The default `pnpm run knip` happens to pass, but `pnpm exec knip --include files` reports:
```
apps/server/tests/setup.ts
```
and exits 1. So a genuinely-used file is one config tweak away from breaking CI. Fix by adding `*.config.ts` to the server workspace `project` (matching electron) so `vitest.config.ts` is parsed, or by adding `tests/setup.ts` as a server entry. The asymmetry between the two workspaces' test globs reads accidental either way.

Neither is a correctness blocker for the feature, but #1 is a 2-line cleanup and #2 makes the gate robust against a false positive on a file that's actually in use.

siddharthkul:
Addressed in `527586e`:

- Removed `node:sqlite` from both Knip `ignoreDependencies` lists.
- Added `*.config.ts` to the server workspace project globs so `vitest.config.ts` is part of the graph.
- Added a Vitest-specific entry override for `tests/setup.ts`, which makes the reviewer’s `--include files` probe pass without a redundant generic entry.
- Pinned the new Knip CI job to Node 22 so `node:sqlite` is resolved as a builtin, matching the server engine and the review environment.

Verified with:
- `npx -y node@22 node_modules/knip/bin/knip.js`
- `npx -y node@22 node_modules/knip/bin/knip.js --include files`
- `corepack pnpm biome check knip.jsonc .github/workflows/build.yml`

MathurAditya724 (approved):
all four points addressed cleanly. the vitest plugin entry for `tests/setup.ts` is a nicer approach than the generic entry i suggested — more semantically correct. lgtm

MathurAditya724 (changes requested):
config looks solid overall — the workspace entries, ignored binaries, and export-rule deferral all make sense. one thing to fix:

the knip job hardcodes `node-version: "22"` while every other job in this workflow uses `node-version-file: "package.json"`. this means the knip job could silently drift to a different Node version than lint, typecheck, tests, and builds. should use `node-version-file: "package.json"` for consistency.

minor note: `packages/sdk`, `packages/utils`, and `packages/validations` aren't listed in the `workspaces` section of `knip.jsonc`, but since knip v6 auto-discovers pnpm workspaces and those packages have standard `src/index.ts` entry points, the defaults should cover them. just confirming that's intentional — if so, no change needed.

siddharthkul:
Addressed in `f7e5c5e`: the Knip job now uses `node-version-file: "package.json"` like the rest of the workflow, so it won't drift from the other CI jobs.

Also confirming the package workspaces omission in `knip.jsonc` is intentional: Knip v6 auto-discovers the pnpm workspaces, and `packages/sdk`, `packages/utils`, and `packages/validations` all fit the default package entry shape.

Verified with:
- `npx -y node@22 node_modules/knip/bin/knip.js`
- `npx -y node@22 node_modules/knip/bin/knip.js --include files`
- `corepack pnpm biome check .github/workflows/build.yml knip.jsonc`

MathurAditya724 (approved):
verified all entry points, ignored files, and workspace paths against the repo — everything checks out. knip passes locally with zero findings, CI is green, and the workflow integration (parallel job gated at `ci-status`) is clean.

one minor nit: the `plugins/*` workspace config omits `*.config.ts` from its `project` glob, unlike `apps/electron` and `apps/server`. the `vite.config.ts` files in each plugin should still be picked up by knip's built-in vite plugin, but adding `"*.config.ts"` to the `plugins/*` project array would keep things consistent. not blocking on it though.

PR #360, opened 2026-06-30
ci: skip expensive builds for unaffected PRs

siddharthkul:
## Summary
- Adds a change-detection job to classify PR paths before running expensive CI jobs.
- Skips Electron platform builds, Electron E2E, and server image builds when a PR does not touch their relevant app/server/shared paths.
- Adds targeted plugin package builds for plugin/SDK changes and keeps full CI behavior on main/release branch pushes.
- Treats `.npmrc` as install-sensitive so pnpm configuration changes still run the affected build paths.
- Uses a merge-base PR diff range for path detection so rebased branches do not over-trigger jobs from unrelated `main` changes.
- Updates Knip config for Node's built-in `node:sqlite` imports after rebasing onto the new Knip CI check.

## Test plan
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/build.yml"); puts "workflow yaml parsed"'`
- Simulated the path classifier for `.npmrc`, docs-only, plugin, server, and workflow changes.
- `corepack pnpm turbo build --filter=@freestyle-voice/plugin-audio-transcription --filter=@freestyle-voice/profanity-filter --dry=json`
- `corepack pnpm run knip`
- `git diff --name-only origin/main...HEAD`
- `git diff --check`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,8 +15,86 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  changes:
+    name: Detect changed areas
+    runs-on: ubuntu-latest
+    outputs:
+      run_quality: ${{ steps.paths.outputs.run_quality }}
+      run_electron: ${{ steps.paths.outputs.run_electron }}
+      run_server_image: ${{ steps.paths.outputs.run_server_image }}
+      run_plugins: ${{ steps.paths.outputs.run_plugins }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Classify changed paths
+        id: paths
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          if [[ "${{ github.event_name }}" != "pull_request" ]]; then
+            {
+              echo "run_quality=true"
+              echo "run_electron=true"
+              echo "run_server_image=true"
+              echo "run_plugins=true"
+            } >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          git diff --name-only \
+            "${{ github.event.pull_request.base.sha }}" \
+            HEAD > changed-files.txt
+
+          run_quality=false
+          run_electron=false
+          run_server_image=false
+          run_plugins=false
+
+          echo "Changed files:"
+          while IFS= read -r file; do
+            echo "- $file"
+
+            case "$file" in
+              docs/*|*.md)
+                ;;
+              *)
+                run_quality=true
+                ;;
+            esac
+
+            case "$file" in
+              apps/electron/*|apps/server/*|packages/*|scripts/*|package.json|pnpm-lock.yaml|pnpm-workspace.yaml|turbo.json|.github/workflows/build.yml)
+                run_electron=true
+                ;;
+            esac
+
+            case "$file" in
+              apps/server/*|packages/sdk/*|packages/utils/*|packages/validations/*|package.json|pnpm-lock.yaml|pnpm-workspace.yaml|turbo.json|.github/workflows/build.yml)
+                run_server_image=true
+                ;;
+            esac
+
+            case "$file" in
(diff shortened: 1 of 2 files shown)


MathurAditya724 (commented):
Reviewed the conditional logic carefully — the structural design is sound. I verified the parts that usually break path-filtered CI and they hold up:

- `ci-status` lists all 10 jobs **directly** in `needs`, so dropping the `skipped` guard does **not** mask real failures (a failing job still surfaces as `failure`). This is the correct pattern, not the usual footgun.
- `run_server_image` implies `run_electron`, and `run_plugins` implies `run_quality`, so there's no path where a build runs while its own quality gate (lint/typecheck) was skipped.
- The plugin turbo filters resolve to real tasks (`pnpm turbo build --filter=... --dry=json` lists both plugin builds + the SDK dep).
- `set -euo pipefail` means a broken `git diff` fails the `changes` job, which `ci-status` catches — safe failure mode.

A few things to address:

**1. `.npmrc` isn't in any path list — this is an under-running gap (the dangerous direction).**
`.npmrc` controls hoisting (`public-hoist-pattern[]=react/react-dom`) and the registry, so it directly affects every `pnpm install --frozen-lockfile`. A PR touching only `.npmrc` classifies as nothing → all builds skip → a change that breaks installs merges green. Add `.npmrc` to at least the `run_electron` and `run_server_image` globs (alongside `pnpm-lock.yaml`/`pnpm-workspace.yaml`).

**2. Docs-only PRs now skip `lint` entirely, which drops Biome from markdown/JSON.**
`biome.json` has `includes: ["**", ...]`, so `pnpm biome check .` currently checks `.md`/`.json` formatting. With `run_quality` false for docs-only PRs, a malformed `README.md`/`docs/*.md` (or any JSON) no longer gets caught. Either accept that tradeoff explicitly, or keep a lightweight `biome check` running on all PRs.

**3. `build-server-image` silently dropped `typecheck` from `needs` (was `[lint, typecheck, test-server]`, now `[changes, lint, test-server]`).**
This isn't in the PR description. It's *safe* in practice because `run_server_image` implies `run_electron`, so `typecheck` still runs in parallel and still gates via `ci-status` — but the image job itself no longer waits on it, so on PRs the image can go green while typecheck is still in flight. If intentional, please note it in the description; if accidental, restore `typecheck`.

**4. (optional) `git diff "base.sha" HEAD` uses the base-branch tip, not the merge-base.**
When `main` advances past the branch point, this over-includes files changed on `main` and runs more jobs than necessary — safe direction, just less precise. `git diff "base.sha...HEAD"` (three-dot, merge-base) would tighten it.

None of these block the feature; #1 is the one I'd want fixed before merge since it can let a build-breaking change through.

siddharthkul:
Thanks for the review. I pushed follow-ups for the actionable items:

- Added `.npmrc` to the install-sensitive path gates so pnpm config changes now trigger Electron, server image, and plugin package validation.
- Switched PR path detection to merge-base semantics and fixed the range syntax to pass it as a single `git diff` argument: `base...HEAD`.
- After rebasing onto latest `main`, Knip started flagging Node's built-in `node:sqlite` imports, so I added the minimal Knip ignores for the Electron/server workspaces.

Validation so far:
- `pnpm run knip` passes locally and in CI.
- `Detect changed areas`, lint, typecheck, server tests, Electron E2E, plugin build, and server image checks are green on the latest run.
- Only the long platform packaging jobs are still pending.

I left docs-only lint skipping as an intentional CI-time tradeoff for now; happy to revisit if we want Biome to stay mandatory for Markdown/docs-only PRs.

MathurAditya724 (approved):
logic looks sound — the path classifier correctly gates each job, dependency chains hold (e.g. `run_electron=true` always implies `run_quality=true` and `test-server` runs), and removing `skipped` from `ci-status` is the right fix now that jobs are conditionally skipped.

a couple things i verified:
- bash `case` patterns like `apps/electron/*` match nested paths (unlike filename globbing), so deep paths are covered
- `build-server-image` correctly drops the `typecheck` dependency since it's electron-specific
- the `node:sqlite` knip fix is unrelated but fine to bundle

nice cleanup merging the two `apt-get` steps in `test-electron` too.

PR #361, opened 2026-06-30
feat(electron): macOS glass/vibrancy window effect

siddharthkul:
## Summary

Adds a Cursor/Codex-style "glass" look to the Electron app **on macOS**: a translucent, blurred sidebar + topbar layered over the native window vibrancy, with a solid main content pane for readability. Non-macOS platforms are unchanged (solid, opaque window).

Screenshot 2026-07-01 at 12 27 36 AM

Screenshot 2026-07-01 at 12 27 19 AM

- **Native vibrancy (macOS only)** — the settings/dashboard `BrowserWindow` enables `transparent`, `vibrancy: "under-window"`, and `visualEffectState: "active"` only when `process.platform === "darwin"`, so the system blur shows through. Windows/Linux windows stay opaque.
- **Layering fix** — the window shell is a transparent, blur-only container. The main content (`glass-content`) is the solid surface and the sidebar (`glass-sidebar`) is the genuinely glassy one. Previously a shared shell background was painted behind everything, so the sidebar always stacked into looking *more* solid than the main pane — the opposite of the intended hierarchy. The backdrop blur lives only on the shell to avoid redundant nested `backdrop-filter`.
- **Frosted surfaces** — the active sidebar nav item and the Freestyle Cloud promo card use translucent frosted styles (`glass-nav-active`, `glass-card`) instead of an opaque `bg-card`, so they blend with the glass rather than reading as hard rectangles.
- **Platform-gated translucency** — surfaces are **solid by default**; the translucent treatment is opt-in via an `html.glass` class added in the renderer entry (`dashboard.tsx`) only when `window.api.platform === "darwin"`. This prevents a regression where, on opaque (non-mac) windows, translucent surfaces would composite over the default white window background and wash out the theme. The solid fallbacks reuse the existing theme tokens (`--background`, `--sidebar`, `--card`, `--border`), so non-mac renders pixel-identical to before this PR.
- **Theme-aware + fallbacks** — light/dark variables for all glass tints, solid `background` fallbacks, and a `prefers-reduced-transparency` path (scoped to the glass path) that drops macOS back to solid surfaces and clears the blur.

## Testing

Tested on macOS (Apple Silicon) during development — verified the native backdrop blur, sidebar/main glass hierarchy, active nav item, and cloud card across light/dark themes.

- [x] macOS: desktop/wallpaper blurs through the sidebar and topbar while the main content stays solid and readable.
- [x] Light and dark themes both look correct.
- [x] Active nav item and cloud card read as frosted (not opaque); active item tuned so it isn't too bright.
- [ ] Onboarding window renders with the blur + solid content.
- [ ] macOS System Settings → Reduce transparency: confirm surfaces fall back to solid.
- [ ] **Windows/Linux: confirm the window is opaque and the UI is unchanged (no `html.glass`, solid surfaces, no washed-out panels).** _(not tested locally — no non-mac machine available)_

### DIFF
--- a/apps/electron/src/renderer/src/onboarding.tsx
+++ b/apps/electron/src/renderer/src/onboarding.tsx
@@ -730,7 +730,7 @@ export default function OnboardingPage(): React.JSX.Element {
   }, [localSetupModel, downloadLocalModel]);
 
   return (
-    <div className="bg-background flex h-screen flex-col">
+    <div className="glass-window-shell glass-content flex h-screen flex-col">
       {!isFullscreen && (
         <div
           className="h-9 shrink-0"
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -574,6 +574,14 @@ function createSettingsWindow(initialPath?: string): void {
     minHeight: 480,
     show: false,
     autoHideMenuBar: true,
+    ...(process.platform === "darwin"
+      ? {
+          backgroundColor: "#00000000",
+          transparent: true,
+          vibrancy: "under-window" as const,
+          visualEffectState: "active" as const,
+        }
+      : {}),
     titleBarStyle: process.platform === "darwin" ? "hiddenInset" : "default",
     trafficLightPosition:
       process.platform === "darwin" ? { x: 16, y: 16 } : undefined,
--- a/apps/electron/src/renderer/src/components/cloud-profile.tsx
+++ b/apps/electron/src/renderer/src/components/cloud-profile.tsx
@@ -30,7 +30,7 @@ export function CloudProfileButton(): React.JSX.Element {
 
   if (!user) {
     return (
-      <div className="border-border bg-card rounded-[10px] border p-3">
+      <div className="glass-card rounded-[10px] border p-3">
         <div className="flex items-center gap-1.5">
           <Cloud className="text-primary size-3.5 shrink-0" />
           <span className="text-foreground text-[12.5px] font-medium">
--- a/apps/electron/src/renderer/src/dashboard.tsx
+++ b/apps/electron/src/renderer/src/dashboard.tsx
@@ -52,6 +52,14 @@ function PagePad(): React.JSX.Element {
 initApiBase();
 installGlobalErrorHandlers();
 
+// Opt into the translucent "glass" surfaces only on macOS, where the window is
+// transparent and backed by native vibrancy. On other platforms the window
+// stays opaque, so surfaces remain solid (see globals.css). Set synchronously
+// before the first paint to avoid a flash of the wrong background.
+if (window.api?.platform === "darwin") {
+  document.documentElement.classList.add("glass");
+}
+
 createRoot(document.getElementById("root")!).render(
   <StrictMode>
     <ErrorBoundary>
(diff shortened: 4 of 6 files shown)


apps/electron/src/renderer/src/onboarding.tsx
```diff
@@ -730,7 +730,7 @@ export default function OnboardingPage(): React.JSX.Element {
   }, [localSetupModel, downloadLocalModel]);
 
   return (
-    <div className="bg-background flex h-screen flex-col">
```
matteo8p, line 733: @siddharthkul Thanks for putting in this PR. Do you know what the consequence of removing existing properties like bg-background will do on other non-mac platforms?

matteo8p (commented):
Looks pretty good. Not concerned about performance implications here since it's just stylistic changes. I do wonder what the implication is on other platforms that are non-Mac.

Happy to approve and merge once that is addressed.

apps/electron/src/renderer/src/onboarding.tsx
siddharthkul: Good callout — addressed in 348da2d. Surfaces are now **solid by default** and the translucent treatment is opt-in via an `html.glass` class that's only added in the renderer entry when `window.api.platform === "darwin"`.

So on non-mac platforms `html.glass` is never present and these classes fall back to their solid rules:
- `glass-window-shell` / `glass-content` → `var(--glass-window-solid)`, which is defined identically to `--background` (`#F4F0E4` light / `#16140F` dark)
- `glass-sidebar` → `--sidebar`, `glass-topbar` → `--background`
- `glass-nav-active` / `glass-card` → `--card` + `--border`

i.e. removing `bg-background` here is a no-op off macOS — it renders pixel-identical to before this PR. The native `transparent` / `vibrancy` flags are also darwin-gated, so non-mac windows stay fully opaque.

Verified by construction (matched the solid fallbacks to the existing theme tokens); I don't have a Windows/Linux machine to eyeball, so I left that item in the test plan.

siddharthkul:
Thanks for the review @matteo8p! The non-mac concern is addressed in 348da2d.

**TL;DR:** the glass treatment is now macOS-only and non-mac renders pixel-identical to before this PR.

- Surfaces are **solid by default**; the translucent look is opt-in via an `html.glass` class added in the renderer entry only when `window.api.platform === "darwin"`.
- Off macOS, the `glass-*` classes fall back to solid rules that reuse the existing theme tokens (`glass-window-shell`/`glass-content` → `--background`, `glass-sidebar` → `--sidebar`, `glass-nav-active`/`glass-card` → `--card`/`--border`), so removing `bg-background` etc. is a no-op there.
- The native `transparent` / `vibrancy` / `visualEffectState` flags are also darwin-gated, so Windows/Linux windows stay fully opaque.

Agree on perf — single `backdrop-filter` layer on the (static-backdrop) window shell, and I removed the redundant nested filters, so scrolling content on top doesn't re-blur.

Only caveat: I don't have a Windows/Linux machine to eyeball, so I left that item unchecked in the test plan — verified by construction against the theme tokens. Let me know if you'd like anything else before merge!

matteo8p (approved):
Ship it

PR #367, opened 2026-07-02
fix(electron): normalize transcription pill bars to the voice band

### LINKED ISSUE #366: Normalize the frequency bar height of the transcription pill.
# Overview

When you use Freestyle to transcribe and the pill appears, the pill shows the loudness of the audio coming in as well as the frequency. Right now, the bar is very skewed to the left, meaning only the low frequency bars appear to have any height. We want to normalize the bars to human speaking frequency. That way, more bar height is shown in the center. 

Image

# How to fix

1. Finding the codebase where the pill is rendered and the bar height is configured.

```
Bar heights / frequency mapping — the live visualizer is inline in apps/electron/src/renderer/src/pages/app.tsx (not components/ui/bar-visualizer.tsx, which is unused/legacy). Key spots:
- Analyser setup, ~line 541: fftSize = 256, smoothingTimeConstant = 0.4 → 128 bins.
- Bin→bar mapping, ~lines 470–482: BARS = 14, sliceSize = floor(128/14) = 9, bar i = the average of bins [i*9, i*9+9).
- Pixel scaling, applyBarsToSvg ~line 440.
``` 

2. The mapping is linear across the full 0–24 kHz spectrum. At 48 kHz each bin is ~187 Hz wide, so bar 0 alone covers ~0–1.7 kHz — which is where nearly all speech energy sits. Bars 2–13 map to 3–24 kHz, where voices have almost nothing, so they stay flat. The tiny fftSize = 256 makes the resolution too coarse to spread it out.

3. The fix is to bump fftSize to 1024/2048 and replace the linear slice with a log-spaced mapping restricted to the voice band (~80 Hz–4 kHz), so energy distributes across the bars.

# First time contributing? 

<p align="left">
</p>

rodriguescarson:
Closes #366

## Problem

The live visualizer on the transcription pill maps FFT bins linearly across the full 0-Nyquist range into 14 bars (`app.tsx`, listening branch). At 48 kHz with `fftSize = 256` (~187 Hz/bin), the first bar alone covers ~0-1.7 kHz, which is where nearly all speech energy sits, so bars 2-13 stay flat and the pill looks skewed hard to the left (as in the issue screenshot).

## Fix

- Bump `fftSize` from 256 to 2048 (~23 Hz/bin, 1024 bins) for enough resolution to spread the voice band across the bars.
- Replace the linear `frequencyBinCount / BARS` slicing with a **log-spaced mapping over the human voice band (80 Hz-4 kHz)**: each bar averages the bins whose frequency falls in its log-spaced sub-range, so bar height is distributed across the pill for normal speech.

`freqDataRef` already sizes itself from `analyser.frequencyBinCount`, so the larger FFT needs no other change. The **volume / IPC audio-level calculation is unchanged** (still the mean of the bar values × 2.5) — this PR only changes how bins map to bar heights, not loudness.

## Verification

- `pnpm --filter @freestyle-voice/electron typecheck:web` (tsc) passes.
- `biome check` clean on the changed file (and formatted with the repo's Biome config).

The visual result is best confirmed by running `pnpm dev` and speaking (the bars now rise across the middle of the pill rather than only the leftmost one); happy to tweak the band edges (80 Hz-4 kHz) or `fftSize` if you'd prefer a different balance.

### DIFF
--- a/apps/electron/src/renderer/src/pages/app.tsx
+++ b/apps/electron/src/renderer/src/pages/app.tsx
@@ -11,6 +11,12 @@ import {
 import { SETTINGS_KEYS } from "../../../shared/settings-keys";
 
 const BARS = 14;
+// Human-voice band used to spread the visualizer bars (#366). Mapping FFT bins
+// linearly across the full 0–Nyquist range piled almost all energy into the
+// first bar, since speech sits below a few kHz; bars are log-spaced over this
+// band instead so height is distributed across them.
+const VOICE_MIN_HZ = 80;
+const VOICE_MAX_HZ = 4000;
 const RISE = 0.55;
 const FALL = 0.22;
 const SVG_WIDTH = 117;
@@ -468,14 +474,25 @@ export default function AppPage(): React.JSX.Element {
       const dataArray = freqDataRef.current;
       if (analyser && dataArray) {
         analyser.getByteFrequencyData(dataArray);
-        const sliceSize = Math.floor(analyser.frequencyBinCount / BARS);
+        const binCount = analyser.frequencyBinCount;
+        const nyquist = analyser.context.sampleRate / 2;
+        const freqToBin = (hz: number) =>
+          Math.min(
+            binCount - 1,
+            Math.max(0, Math.round((hz / nyquist) * (binCount - 1))),
+          );
+        const ratio = VOICE_MAX_HZ / VOICE_MIN_HZ;
         const raw: number[] = [];
         let totalSum = 0;
         for (let i = 0; i < BARS; i++) {
+          const loBin = freqToBin(VOICE_MIN_HZ * ratio ** (i / BARS));
+          const hiBin = Math.max(
+            loBin + 1,
+            freqToBin(VOICE_MIN_HZ * ratio ** ((i + 1) / BARS)),
+          );
           let sum = 0;
-          for (let j = 0; j < sliceSize; j++)
-            sum += dataArray[i * sliceSize + j];
-          const val = sum / sliceSize / 255;
+          for (let j = loBin; j < hiBin; j++) sum += dataArray[j];
+          const val = sum / (hiBin - loBin) / 255;
           raw.push(val);
           totalSum += val;
         }
@@ -538,7 +555,9 @@ export default function AppPage(): React.JSX.Element {
 
       const source = ctx.createMediaStreamSource(stream);
       const analyser = ctx.createAnalyser();
-      analyser.fftSize = 256;
+      // 2048 → 1024 bins (~23 Hz each at 48 kHz), enough resolution to
+      // log-spread the voice band across the bars (#366).
+      analyser.fftSize = 2048;
       analyser.smoothingTimeConstant = 0.4;
       source.connect(analyser);
       audioSourceRef.current = source;


matteo8p:
@rodriguescarson Hey Carson, I do really appreciate you putting in this PR. We do have a policy that if you do want to take on a GitHub issue, please comment first and have it be assigned. I did assign this GitHub issue to another person who had asked.

Would you be interested in taking on the other GitHub issue? I saw you open up another PR. If you're interested in that GitHub issue, please drop a comment there.

apps/electron/src/renderer/src/pages/app.tsx
```diff
@@ -468,14 +474,25 @@ export default function AppPage(): React.JSX.Element {
...
+            freqToBin(VOICE_MIN_HZ * ratio ** ((i + 1) / BARS)),
+          );
           let sum = 0;
-          for (let j = 0; j < sliceSize; j++)
-            sum += dataArray[i * sliceSize + j];
-          const val = sum / sliceSize / 255;
+          for (let j = loBin; j < hiBin; j++) sum += dataArray[j];
+          const val = sum / (hiBin - loBin) / 255;
           raw.push(val);
           totalSum += val;
```
MathurAditya724, line 497: Quick flag on the "volume / IPC level is unchanged" line in the description — the formula is the same, but I think the actual reading shifts here. Before, bars 1–13 were basically flat (that was the bug), so `totalSum` got dragged way down and the `* 2.5` gain was effectively tuned against that near-zero padding. Now every bar sits inside the voice band and carries real energy, so `totalSum / BARS` comes out noticeably bigger and `volume` will hit the `Math.min(1, …)` ceiling a lot more often.

That level feeds the onboarding tutorial's live wave (`sendAudioLevel` here → `onAudioLevel` in `tutorial-demo.tsx`), so the wave might peg near-max while you're talking. I don't think it's wrong — the old reading was probably too quiet anyway — but it's a real behavior change. Could we either re-tune the `2.5` for the new distribution, or just soften the PR note so it doesn't claim "unchanged"? Did you get a chance to eyeball the level during the tutorial flow?

MathurAditya724 (commented):
Took this for a spin and read through the whole thing — really nice fix. The log-spaced mapping over the voice band is exactly the right call, the diagnosis in the description matches what I see, and the diff stays tight (sample-rate aware via `analyser.context.sampleRate`, `Math.max(loBin + 1, …)` keeps you safe from a zero-width bar, etc.). 

One thing I want to make sure we're aligned on before it lands — left an inline note about the volume/IPC level. And a tiny nit: `nyquist`, `ratio` and the `freqToBin` closure get rebuilt every RAF frame even though they only depend on `binCount`/`sampleRate`. It's cheap so totally fine to leave, but they'd sit nicely next to the analyser setup in `startListening` if you ever want to hoist them.

Nothing blocking from my side once the volume bit is sorted out.

matteo8p:
@MathurAditya724 The normalized transcription pill work is going to this [PR first](https://github.com/freestyle-voice/freestyle/pull/376). This is because Ayush claimed the issue first. 

If the PR with Ayush stalls, then I will consider this PR and come back to it. Hopefully that makes sense!

matteo8p:
Closing out this PR. Ayush's fix for this was merged. 

https://github.com/freestyle-voice/freestyle/pull/376

PR #368, opened 2026-07-02
fix(electron): soften the transcription pill start/stop tone

### LINKED ISSUE #365: Make the "pop" audio of the transcription pill a softer sound
# Overview 

Right now, when you hit the hotkey and the pill appears to start transcribing, Freestyle creates a pop indicator sound. This pop also happens when you release the hotkey and finish transcription. 

Right now the pop is pretty aggressive and harsh. We want to make the pop a lot softer. 

Image

# How to fix

1. Right now, the pop sound is a sine-wave tone synthesized at runtime with the Web Audio API. 

```
apps/electron/src/renderer/src/pages/app.tsx
- playTone() player — lines ~30–60 (shared AudioContext → OscillatorNode + GainNode with exponential decay)
- TONE_PRESETS — lines ~37–41 (start: 880 Hz / 100 ms, stop: 660 Hz / 100 ms)
- Appears → playTone("start") at line ~650
- Disappears → playTone("stop") at line ~741
- Gated by _soundEnabled, loaded from the soundEnabled setting (toggle in settings.tsx, copy in locales/en.json:164)
```

2. Use the coding agent and prompt the coding agent to make the sound softer. Or find an audio file to import into the freestyle project and use that instead. 

# First time contributing? 

<p align="left">
</p>

rodriguescarson:
Closes #365

## Problem

The start/stop indicator tone (`playTone` in `app.tsx`) jumped straight to full gain at onset:

```ts
gain.gain.setValueAtTime(volume, ctx.currentTime); // instant -> click
```

That instant onset is the main source of the harsh "pop", and the 880/660 Hz sines are bright on top of it.

## Fix

- **Fade the tone in** over ~20 ms (`linearRampToValueAtTime`) instead of jumping to full volume. This removes the onset click, which is the biggest part of the harshness.
- **Lower the default volume** 0.3 → 0.2.
- **Warm the frequencies**: D5 (587 Hz) / A4 (440 Hz) instead of 880 / 660 Hz, keeping start higher than stop so the two are still distinguishable.
- **Lengthen the envelope** 100 → 140 ms so it reads as a gentle blip rather than a sharp pop.

Still a runtime-synthesized tone (no new audio asset), gated by the same `_soundEnabled` setting.

## Verification

- `pnpm --filter @freestyle-voice/electron typecheck:web` passes.
- `biome check` clean on the changed file.

The exact voicing is subjective, so the frequencies / volume / attack are easy to tune. Happy to adjust to taste, or switch to an imported audio file instead if you'd prefer that direction.

### DIFF
--- a/apps/electron/src/renderer/src/pages/app.tsx
+++ b/apps/electron/src/renderer/src/pages/app.tsx
@@ -35,18 +35,19 @@ function getToneCtx(): AudioContext {
 }
 
 type TonePreset = "start" | "stop";
-// Warmer, lower tones than the original 880/660 Hz, with a longer envelope, so
-// the start/stop indicator is a gentle blip rather than a harsh pop (#365).
+// Low, quiet tones (an octave below the first pass at D5/A4, two below the
+// original 880/660 Hz) so the start/stop indicator is a soft low blip in the
+// Wispr Flow vein rather than a harsh pop (#365).
 const TONE_PRESETS: Record<TonePreset, { freq: number; ms: number }> = {
-  start: { freq: 587.33, ms: 140 }, // D5
-  stop: { freq: 440, ms: 140 }, // A4
+  start: { freq: 293.66, ms: 170 }, // D4
+  stop: { freq: 220, ms: 170 }, // A3
 };
 
 // Fade the tone in over a few ms instead of jumping to full volume instantly.
 // The instant onset was the main source of the harsh "click" in the pop.
-const TONE_ATTACK_S = 0.02;
+const TONE_ATTACK_S = 0.03;
 
-async function playTone(preset: TonePreset, volume = 0.2): Promise<void> {
+async function playTone(preset: TonePreset, volume = 0.12): Promise<void> {
   if (!_soundEnabled) return;
   const { freq, ms } = TONE_PRESETS[preset];
   try {
--- a/apps/electron/src/renderer/src/pages/app.tsx
+++ b/apps/electron/src/renderer/src/pages/app.tsx
@@ -35,12 +35,18 @@ function getToneCtx(): AudioContext {
 }
 
 type TonePreset = "start" | "stop";
+// Warmer, lower tones than the original 880/660 Hz, with a longer envelope, so
+// the start/stop indicator is a gentle blip rather than a harsh pop (#365).
 const TONE_PRESETS: Record<TonePreset, { freq: number; ms: number }> = {
-  start: { freq: 880, ms: 100 },
-  stop: { freq: 660, ms: 100 },
+  start: { freq: 587.33, ms: 140 }, // D5
+  stop: { freq: 440, ms: 140 }, // A4
 };
 
-async function playTone(preset: TonePreset, volume = 0.3): Promise<void> {
+// Fade the tone in over a few ms instead of jumping to full volume instantly.
+// The instant onset was the main source of the harsh "click" in the pop.
+const TONE_ATTACK_S = 0.02;
+
+async function playTone(preset: TonePreset, volume = 0.2): Promise<void> {
   if (!_soundEnabled) return;
   const { freq, ms } = TONE_PRESETS[preset];
   try {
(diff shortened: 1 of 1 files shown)


matteo8p:
@rodriguescarson Hey Carson, really appreciate you putting out this PR. Would you be willing to leave a comment in the parent's GitHub issue? That way I can assign it to you. We do have a policy where people have to be assigned a club issue before they work on it. That way we prevent conflicts and duplicated work.

apps/electron/src/renderer/src/pages/app.tsx
```diff
@@ -50,12 +56,16 @@ async function playTone(preset: TonePreset, volume = 0.3): Promise<void> {
...
+    const attack = Math.min(TONE_ATTACK_S, dur * 0.3);
+    gain.gain.setValueAtTime(0.0001, now);
+    gain.gain.linearRampToValueAtTime(volume, now + attack);
+    gain.gain.exponentialRampToValueAtTime(0.0001, now + dur);
     osc.connect(gain);
     gain.connect(ctx.destination);
-    osc.start();
-    osc.stop(ctx.currentTime + ms / 1000);
+    osc.start(now);
+    osc.stop(now + dur);
```
MathurAditya724, line 69: This is the heart of the fix and it's spot on — ramping up to `volume` over ~20 ms instead of the old instant `setValueAtTime(volume, now)` is what removes the onset click. The `setValueAtTime(0.0001, now)` seed + `linearRampToValueAtTime` → `exponentialRampToValueAtTime(0.0001, …)` chain is all valid (non-zero targets keep the exponential ramp legal), and the tail reaches near-silence right as `osc.stop` fires, so no click on the way out either. Nice.

MathurAditya724 (commented):
Read through it and gave the logic a close look — this is a really clean fix. The attack fade-in is the part that actually matters: the old `setValueAtTime(volume, now)` jumped straight to full gain, and that instant onset is exactly what makes the click. Swapping it for a short linear ramp is the textbook way to kill that transient, so you nailed the root cause rather than just papering over it.

A few things I liked while going through it:
- The envelope scheduling is correct — non-zero targets so the exponential ramp stays legal, and the tail decays to near-silence before `osc.stop`, so there's no closing click either.
- `attack = Math.min(TONE_ATTACK_S, dur * 0.3)` is a nice guard so the fade-in can never swallow the whole envelope if `ms` ever shrinks.
- Both call sites use the default volume and nothing overrides it, so the 0.3 → 0.2 change lands uniformly and the duck/restore ordering around the tones is untouched.

One subjective thought, not blocking: you're softening in four directions at once — quieter, lower/warmer pitch, a fade-in, and a longer decay. Each one on its own tames the harshness, so stacking all four might tip it toward "too subtle to notice," especially the 440 Hz stop tone on laptop speakers with weak low-mids. Worth a quick listen on real hardware just to confirm it still reads as a clear start/stop cue and not just quiet. Totally fine to leave as-is if it sounds good to you — the mechanism is right either way.

rodriguescarson:
Thanks for the close read @MathurAditya724! On the stacking concern — fair point. The only part that's load-bearing for the click is the 20 ms attack ramp; the lower pitch (D5/A4), 0.2 gain, and 140 ms tail are taste, aiming for the "gentle blip" asked for in #365. If the stop tone reads too subtle on laptop speakers, volume is a free knob — the ramp kills the click at any gain — so happy to bring the default back to 0.25–0.3, or trim the tail to ~120 ms. Whichever direction sounds right to you both.

@matteo8p thanks for assigning #365 — and I've joined the Discord 👋

matteo8p:
> Thanks for the close read @MathurAditya724! On the stacking concern — fair point. The only part that's load-bearing for the click is the 20 ms attack ramp; the lower pitch (D5/A4), 0.2 gain, and 140 ms tail are taste, aiming for the "gentle blip" asked for in #365. If the stop tone reads too subtle on laptop speakers, volume is a free knob — the ramp kills the click at any gain — so happy to bring the default back to 0.25–0.3, or trim the tail to ~120 ms. Whichever direction sounds right to you both.
> 
> @matteo8p thanks for assigning #365 — and I've joined the Discord 👋

Hey @rodriguescarson I just reviewed the PR and tried it out myself. I think the new sound is still a little bit too harsh. Have you tried out Wispr Flow's sound? I quite like their sound. It's low pitched and very subtle. Not too harsh.

I think we want something quite similar. More lower pitched so it's not harsh on the ears. Let me know if you think you'd be able to make those changes. Let's stay in touch on Discord. Happy to help with anything.

rodriguescarson:
Pushed an update: both tones are down a full octave now (D4/A3 instead of D5/A4), gain cut from 0.2 to 0.12, slightly slower attack and a longer tail. Aimed for that low soft Wispr Flow feel rather than a chime. Give it a listen and if it should sit even lower or quieter, both are one-line knobs.

rodriguescarson:
Just saw #433 landed and closed #365 with your own take on the tone (F4/C4, softer envelope with a proper release ramp). That covers what this PR was for, so I'll close it. Thanks for the reviews and the Discord invite @matteo8p, happy to pick up another issue.

PR #376, opened 2026-07-02
fix(electron): normalize transcription pill bars to the voice band

AyushSrivastava1818:
## Summary

This PR improves the transcription pill's frequency visualizer by focusing the bar mapping on the human voice frequency range instead of the entire FFT spectrum.

## Changes

- Increased `AnalyserNode` FFT size from 256 to 1024 for better frequency resolution.
- Replaced the linear FFT bin mapping with a logarithmic mapping across the voice band (80 Hz–4 kHz).
- Preserved the existing smoothing and SVG rendering logic.

## Result

The visualizer distributes speech energy more evenly across the bars, making the transcription pill appear more balanced and responsive during speech.

## Testing

- [ ] Tested locally
- [x] Code reviewed

### DIFF
--- a/apps/electron/src/renderer/src/pages/app.tsx
+++ b/apps/electron/src/renderer/src/pages/app.tsx
@@ -474,37 +474,42 @@ export default function AppPage(): React.JSX.Element {
         const sampleRate = analyser.context.sampleRate;
         const binWidth = sampleRate / analyser.fftSize;
 
+        const startBin = Math.max(0, Math.floor(VOICE_MIN / binWidth));
+        const endBin = Math.min(
+          analyser.frequencyBinCount,
+          Math.ceil(VOICE_MAX / binWidth),
+        );
+
+        // Compute one overall voice level
+        let sum = 0;
+        for (let i = startBin; i < endBin; i++) {
+          sum += dataArray[i];
+        }
+
+        const voiceLevel = sum / (Math.max(1, endBin - startBin) * 255);
+
         const raw: number[] = [];
-        let totalSum = 0;
+        const center = (BARS - 1) / 2;
+        const sigma = BARS / 4;
+
+        const binCount = Math.max(1, endBin - startBin);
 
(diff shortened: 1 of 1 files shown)


matteo8p (changes requested):
Hey, I just tested this out. It did not centralize the bar. 

It made all of the bars more sensitive.

AyushSrivastava1818:
Thanks for testing! I think I understand the issue with the current approach—it spreads the speech energy across all bars, so they all become more active, but it doesn't create the centered visual you're looking for.

Just to make sure I implement the intended behavior correctly: when you say "centralize the bar", do you mean the visualization should be centered visually (with the middle bars showing the strongest activity, similar to Siri/ChatGPT Voice), rather than being a direct frequency spectrum?

I want to make sure I match the intended UX instead of just tweaking the FFT mapping again.

matteo8p:
> Just to make sure I implement the intended behavior correctly: when you say "centralize the bar", do you mean the visualization should be centered visually (with the middle bars showing the strongest activity, similar to Siri/ChatGPT Voice), rather than being a direct frequency spectrum?

Sorry for the late response, it was July 4th holiday over the weekend. 

This is exactly what my intention was! The intention is to have the visualization be centered, so the middle bars show the strongest activity. Let me know if you need any help or have any questions!

AyushSrivastava1818:
Thanks for confirming! That makes perfect sense. I'll update the implementation so the visualization is centered, with the middle bars carrying the strongest activity rather than representing a direct frequency spectrum

AyushSrivastava1818:
Updated the implementation based on the review feedback. The visualization is now centered around the middle bars while still using voice-band FFT data to introduce subtle variation across the bars.

matteo8p (approved):
@AyushSrivastava1818 Just tried it out. It works great. Thank you so much for putting this in! 

Also, if you have a Discord, we'd love to have you join it. That's where other project contributors and I hang out. 

https://discord.gg/p3kGY93Kt

PR #377, opened 2026-07-02
fix: support pnpm 11 electron installs

tigerisaac:
pnpm 11 no longer reads package.json -> pnpm.onlyBuiltDependencies. 

Without the new allowBuilds setting, it blocks Electron’s install script, so the Electron binary never gets installed and pnpm dev dies with the “Electron uninstall” / “Electron failed to install correctly” error. Moving the allowlist to pnpm-workspace.yaml lets electron and esbuild run their required postinstall scripts again


matteo8p (approved):
Ship it

PR #378, opened 2026-07-02
fix(electron): improve light mode macOS glass readability

siddharthkul:
## Summary
- Makes the main content pane fully solid in light mode so colorful wallpapers no longer bleed through the reading area
- Raises sidebar/chrome opacity and uses the correct sidebar tint (`#ECE7D6`) instead of the window cream
- Desaturates the light-mode backdrop blur so frosted surfaces stay warm and clean without muddy color bleed

Dark mode glass tokens and behavior are unchanged.

Screenshot 2026-07-02 at 9 56 19 PM

Screenshot 2026-07-02 at 9 56 38 PM

## Test plan
- [ ] macOS light mode: main content area reads as solid warm paper; sidebar still shows native blur without pink/orange wallpaper bleed
- [ ] macOS dark mode: glass effect unchanged from #361
- [ ] Active nav item and Freestyle Cloud card still read as frosted, not opaque blocks
- [ ] macOS System Settings → Reduce transparency: surfaces still fall back to solid
- [ ] Windows/Linux: no `html.glass` — UI unchanged

### DIFF
--- a/apps/electron/src/renderer/src/globals.css
+++ b/apps/electron/src/renderer/src/globals.css
@@ -90,17 +90,17 @@
   --sidebar-border: #D6CDB8;
   --sidebar-ring: #6B8F12;
   --glass-window-solid: #F4F0E4;
-  --glass-window: rgba(244, 240, 228, 0.82);
+  --glass-window: #F4F0E4;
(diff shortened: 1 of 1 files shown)


matteo8p (approved):
Ship it

PR #379, opened 2026-07-02
feat: voice-commands plugin with LLM agent action routing

MathurAditya724:
## What

Adds a new **`@freestyle-voice/commands`** plugin that turns spoken trigger phrases into actions, plus the two small SDK/server seams it needs.

Say something like _"post to Slack that I'll be five minutes late"_ and — instead of typing the words — Freestyle runs the matching action.

## How it works

1. **Prefilter** — a cheap, deterministic phrase match gates everything. No trigger phrase → nothing else runs and the transcript is dictated normally.
2. **Agent** — when a trigger matches, a multi-step tool-calling agent (using the LLM you've already configured for cleanup) decides whether it's really a command and extracts the payload (the non-trigger part of the utterance).
3. **Action** — the command fires and the utterance is *consumed*: cleanup is skipped and no text is delivered.

If no cleanup LLM is configured, detection falls back to a deterministic match (longest matching trigger wins) so commands still work.

### Action types
- **Run Shortcut** *(macOS only, hidden elsewhere)* — runs a macOS Shortcut by name; payload piped to stdin.
- **Call webhook** — `POST` payload as JSON, or `GET` as a query param.
- **Open URL / app** — opens a URL or app scheme; `{{input}}` substituted.
- **Run script** — runs a shell command; `{{input}}` substituted and payload also in `$FREESTYLE_COMMAND_INPUT`.

## SDK / server changes

- **`afterTranscribe` gains `output.consumed`** — when a plugin sets it, the host skips LLM cleanup and delivers nothing. Honored in both the batch `/transcribe` route and the streaming `/stream` route.
- **`PluginContext.llm`** — a **server-only** capability exposing the user's default cleanup model + stored keys (`getModel()` returns an AI SDK `LanguageModel`). Plugins reuse the host's LLM and never manage their own credentials. Always guard with `if (ctx.llm)`.

Both additions are optional/additive, so existing plugins are unaffected.

## Notes / decisions

- The plugin bundles `ai` + `zod` (in `devDependencies`, à la the profanity-filter UI) because installed plugins don't get a transitive `npm install`. The host-created `LanguageModel` is passed to the plugin's bundled `generateText` — both are AI SDK v6, so the `LanguageModelV2` spec object is compatible across copies.
- `consumed` returns `raw: ""`, which the desktop app already filters out (`nonEmpty`), so no paste occurs — consistent with the existing empty-transcript path.
- The plugin must still be added to the `plugins` allowlist (enabled in the UI) for the server to execute it — same as every other plugin.

## Verification

- `freestyle-voice` (SDK) build + tests (9 passing)
- `@freestyle-voice/commands` typecheck (server + UI) and full build
- `@freestyle-voice/server` build (tsc typecheck)
- `@freestyle-voice/electron` typecheck (node + web)


MathurAditya724:
Superseded by #477. The plugin-pipeline overhaul (#462) landed this PR's two SDK seams in a better form — `afterTranscribe` `output.consumed` → `api.control.consume()`, and `PluginContext.llm` → `api.llm` — so #477 keeps just the plugin, rebuilt against the current SDK. Closing in favor of that.

PR #382, opened 2026-07-03
New tone page to replace formats page

tigerisaac:
This PR replaces the old Formats page with a new Tone page where users can set cleanup intensity and separate tone preferences for personal texts, work messages, email, and everything else. It also adds destination-aware routing, so tones can be mapped to built-in app groups, specific open apps, and individual websites, with route icons and live previews throughout the UI.
Under the hood, it wires the new tone settings through shared validation/settings plumbing, updates server-side post-processing and rewrite-context logic to respect those settings, and syncs tone preferences to Freestyle Cloud, including right after sign-in.


matteo8p:
Thanks for putting in this PR. Just a high-level product question. How is the tone set? Are we using context aware to determine which tone to use?

matteo8p:
@tigerisaac Converting this to a draft right now. For now.

MathurAditya724:
Hey @tigerisaac — folding this into #387 so the tone system ships together with the cloud v2 payload migration. Your tone work is included there (with the sync reworked from the v1 `/preferences` endpoint to the v2 request payload), and you're credited as co-author. Closing this as superseded by #387.

PR #384, opened 2026-07-03
Phase 1 prs for stability improvements

matteo8p:
1. cfb7468 — CI native-build safety. Added the missing
  <sys/ioctl.h> include to the Linux key listener (hard
  compile error on gcc 14+) and fixed its false "XRecord
  fallback" comment; compile-native.js now exits non-zero in
  CI when any binary fails to compile; a new
  verify-native-binaries.mjs asserts every expected binary
  exists in both resources/bin and the electron-builder
  unpacked package, wired into all three platform jobs;
  artifact uploads now use if-no-files-found: error.
  2. 9e9975e — Dictionary engine. Replacement values with $
  patterns are now inserted literally via a function replacer;
  compiled regexes are cached across transcripts; usage
  counts update in a single batched UPDATE ... IN (...);
  key/value length caps added to validation. Four new test
  cases cover the $ patterns, batching, and cache reuse.
  3. 3b4728d — Hotkey resilience. A new onPermanentFailure
  hook fires when the listener exhausts restarts after a
  successful start — main now installs the globalShortcut
  toggle fallback and notifies instead of silently losing
  dictation; a 5-minute stuck-key watchdog force-releases
  hold-mode presses that never see a key-up; the degradation
  notice re-arms after each healthy start; dead updateHotkey()
  removed.
  4. e330124 — Clipboard restore. The prior clipboard is
  snapshotted with all restorable flavors
  (text/HTML/RTF/image) and restored intact; restore is
  skipped when the clipboard holds unreadable formats (file
  copies) or when it changed during the settle window; native
  settle raised to 300 ms (invisible to the user — it only
  delays the restore) with a FREESTYLE_PASTE_SETTLE_MS
  override; paste calls are serialized.
  5. 9ef6b6e — Wayland portal paste. The Wayland chain is now
  uinput → RemoteDesktop portal → wtype, with the portal's
  restore token persisted so only the first paste prompts;
  revoked grants clear the stale token; cross-family fallback
  handles session misdetection; setup checks probe
  /dev/uinput, onboarding stops warning about wtype when
  uinput covers pasting, and GNOME users are no longer told to
  install a tool that can't work for them.
  6. d68df12 — Audio recovery. The pre-duck volume snapshot is
  persisted to disk and recovered on next launch if a run
  died mid-dictation (guarded so a manually-fixed volume is
  never yanked); Linux media resume on quit is now synchronous
  so music actually resumes; the pactl mic subscriber only
  respawns on abnormal exit.

### DIFF
--- a/packages/validations/src/dictionary.ts
+++ b/packages/validations/src/dictionary.ts
@@ -3,7 +3,7 @@ import { z } from "zod/v3";
 export const DICTIONARY_KEY_MAX = 200;
 export const DICTIONARY_VALUE_MAX = 5000;
 
-export const createDictionarySchema = z.object({
+export const dictionarySchema = z.object({
   key: z
     .string()
     .min(1, "Key is required")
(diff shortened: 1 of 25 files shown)


packages/validations/src/dictionary.ts
```diff
@@ -1,13 +1,30 @@
 import { z } from "zod/v3";
 
+export const DICTIONARY_KEY_MAX = 200;
+export const DICTIONARY_VALUE_MAX = 5000;
+
 export const createDictionarySchema = z.object({
```
MathurAditya724: we can have a `dictionarySchema` and a `.partial()` version of this for updates

matteo8p: Dunzo

Issue #364, opened 2026-07-02
Feat: Enterprise/offline model install support for corporate proxy networks


anish808:
# Add enterprise-friendly local model installation support (proxy/coaching page detection, manual import, custom mirrors)

## Summary

I'm using Freestyle on a corporate network with Zscaler / secure web gateway (SWG) controls. The app is approved internally, but first-run local model installation currently fails because model downloads are performed directly from Hugging Face / GitHub release assets.

In this environment, Hugging Face access is behind a 24-hour coaching/click-through page. Browser access works fine after acknowledging the notice, but Freestyle's in-app fetch sees the coaching HTML page (or fails outright) and surfaces a generic `fetch failed` / model install error instead of something actionable.

**This is not a request to send company data to Hugging Face.** The model weights are static files used locally by Freestyle for offline transcription. The problem is purely about first-run model installation on managed corporate networks.

**Example affected source:**
`https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-base-q5_1.bin`

### 1. Detect proxy/coaching HTML responses
If a model download receives `text/html`, an unexpectedly small file, or a captive/coaching page instead of a model binary, show a clear message instead of `fetch failed`:

> "Your network requires browser acknowledgement before downloading from Hugging Face. Open Hugging Face in browser, click Continue, then retry."

### 2. Add an "Open model source in browser" button
For Hugging Face-backed models, let users open the model/source URL in their default browser so corporate click-through/SWG acknowledgement can happen before retrying the in-app download.

### 3. Add manual model import
Let users select a previously downloaded model file, validate expected filename/size/hash, and copy it into Freestyle's model cache. This helps users whose company requires browser-based downloads or approved internal download tooling.

### 4. Support enterprise mirrors / custom model base URLs
Allow a setting or env var for model download base URLs, so companies can mirror Freestyle-compatible model files into an approved internal artifact repository.

### 5. Respect proxy and custom CA configuration
Ensure all Electron/Node model downloads, Hugging Face downloads, updater downloads, and any spawned Python/Hugging Face processes respect:

- `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY`
- `NODE_EXTRA_CA_CERTS` / `SSL_CERT_FILE`
- OS proxy settings, where possible

### 6. Optional offline/bundled installer
Consider an enterprise/offline installer or release flavor that includes at least the small "Whisper Fast" model, so local transcription works without a runtime Hugging Face download.

[...]

PR #388, opened 2026-07-03
feat: enterprise proxy and custom CA support for model downloads

MathurAditya724:
Local model installation fails on managed corporate networks: downloads go direct to Hugging Face / GitHub with no proxy or custom-CA support, and secure-web-gateway interception surfaces only as a bare `fetch failed`. This PR addresses items **1 (proxy/coaching detection)** and **5 (proxy + custom CA)** of #364.

- Installs a global undici dispatcher (`ProxyAgent` / `EnvHttpProxyAgent` / custom-CA `Agent`) so every server `fetch` — model downloads, HF API, cloud API, updater — honors a proxy and extra CA certs. Config resolves from settings first, then `HTTP(S)_PROXY` / `NO_PROXY` / `NODE_EXTRA_CA_CERTS` / `SSL_CERT_FILE`.
- Detects proxy/coaching HTML (or a suspiciously tiny body) during whisper downloads and raises actionable guidance instead of writing a captive page to disk.
- Maps bare connection/TLS failures to guidance that points at the browser-acknowledgement flow and the new Network settings.
- Adds a **Network** settings section (proxy URL + custom CA path), applied live without an app restart.

Out of scope for this PR (remaining #364 items): manual model import (3), custom mirror base URLs (4), bundled offline installer (6). Item 2 ("open source in browser") is partially covered by the actionable error text; a dedicated button can follow.

## Testing
- `pnpm --filter @freestyle-voice/server test` — 200 passed (adds `download-guard.test.ts` + proxy/TLS cases in `disk.test.ts`)
- Server + electron web/node typechecks pass
- undici dispatcher smoke-tested for proxy / env-proxy / custom-CA paths

Refs #364


anish808:
### Great improvement for corporate proxy environments

Tested this on a State Farm-managed Mac behind Zscaler/proxy, and the new behavior is a big improvement.

The PR correctly detected the Zscaler/coaching HTML response during the Whisper Fast download instead of writing the HTML page into the model cache as `ggml-base-q5_1.bin`.

The error message also made the failure understandable and pointed me toward the browser acknowledgement flow / Network settings, which is exactly the right direction.

### UX follow-up suggestion

One UX improvement that would make this even better: could the model download error dynamically expose the actual `sourceUrl` as a clickable action (for example, **"Open model source"**)?

For corporate coaching pages like ours, users have to open the exact Hugging Face model URL in their browser and complete the 24-hour acknowledgement before retrying the download. The current message says *"Open the model source in your browser,"* but most users won't know what the model source URL is, especially since it differs across model options.

It looks like the code already carries the source URL on `ProxyInterceptionError`, so ideally the UI could expose it as a button or link:

- **Open model source**
  - Opens the specific Hugging Face URL for the selected model.
  - User completes the browser acknowledgement.
  - User returns to Freestyle and retries the download.

That would make this PR much closer to friction-free for managed corporate networks while still respecting the proxy/coaching flow.

PR #389, opened 2026-07-04
 fix: consolidate two LLM settings queries into one (#255)

### LINKED ISSUE #255: Consolidate two "fetch LLM settings" queries into one (easy)
# Overview

Right now, the `fetchLocalLlmModels` function runs two separate queries from sqlLite. Instead, they could be compiled into a single db query. 

Where: `apps/server/src/routes/models.ts` (fetchLocalLlmModels, ~lines 41–52)

# How to fix 
1. Replace the two `.get()` calls with one query:
   ```typescript
   const rows = db
     .prepare("SELECT key, value FROM settings WHERE key IN ('local_llm_url', 'local_llm_api_key')")
     .all() as { key: string; value: string }[];
   const settings = Object.fromEntries(rows.map((r) => [r.key, r.value]));
   if (!settings.local_llm_url) return [];
   ```
2. Update the rest of the function to use `settings.local_llm_url` / `settings.local_llm_api_key`.

To verify that everything works, do a functional test to make sure that local models work. Nothing is breaking within Freestyle.

# First time contributing? 

<p align="left">
</p>

Olaiwonismail:
# PR Title
fix: consolidate two LLM settings queries into one

Closes #255

## What

Replaced two separate `db.prepare().get()` calls in `fetchLocalLlmModels` with a single `db.prepare().all()` query using `WHERE key IN (...)`.

## Where

`apps/server/src/routes/models.ts` — `fetchLocalLlmModels` function (lines 46–56)

## Changes

**Before:** Two round-trips to SQLite — one for `local_llm_url`, one for `local_llm_api_key`.

**After:** One query fetches both keys, results mapped via `Object.fromEntries`.

### DIFF
--- a/apps/server/src/routes/models.ts
+++ b/apps/server/src/routes/models.ts
@@ -45,20 +45,23 @@ const UNSUITABLE_CLEANUP_MODEL_PATTERN =
 
 async function fetchLocalLlmModels(): Promise<AvailableModel[]> {
   const db = getDb();
-  const urlRow = db
-    .prepare("SELECT value FROM settings WHERE key = 'local_llm_url'")
-    .get() as { value: string } | undefined;
-  if (!urlRow?.value) return [];
+  const rows = db
+    .prepare(
+      "SELECT key, value FROM settings WHERE key IN ('local_llm_url', 'local_llm_api_key')",
+    )
+    .all() as { key: string; value: string }[];
+  const settings = Object.fromEntries(rows.map((r) => [r.key, r.value]));
+  if (!settings.local_llm_url) return [];
 
-  const keyRow = db
-    .prepare("SELECT value FROM settings WHERE key = 'local_llm_api_key'")
-    .get() as { value: string } | undefined;
-
-  const baseUrl = urlRow.value.replace(/\/+$/, "").replace(/\/v1$/, "");
+  const baseUrl = settings.local_llm_url
+    .replace(/\/+$/, "")
+    .replace(/\/v1$/, "");
 
   const res = await fetch(`${baseUrl}/v1/models`, {
     headers: {
-      ...(keyRow?.value ? { Authorization: `Bearer ${keyRow.value}` } : {}),
+      ...(settings.local_llm_api_key
+        ? { Authorization: `Bearer ${settings.local_llm_api_key}` }
+        : {}),
     },
     signal: AbortSignal.timeout(3000),
   });


apps/server/src/routes/models.ts
```diff
@@ -45,20 +45,23 @@ const UNSUITABLE_CLEANUP_MODEL_PATTERN =
...
-
-  const baseUrl = urlRow.value.replace(/\/+$/, "").replace(/\/v1$/, "");
+  const rows = db
+    .prepare(
+      "SELECT key, value FROM settings WHERE key IN ('local_llm_url', 'local_llm_api_key')",
+    )
+    .all() as { key: string; value: string }[];
+  const settings = Object.fromEntries(rows.map((r) => [r.key, r.value]));
+  if (!settings.local_llm_url) return [];
+
```
MathurAditya724, line 55: Non-blocking: `Object.fromEntries` returns `any`, so `settings.local_llm_url` / `settings.local_llm_api_key` are now unchecked property accesses — a typo in either key would compile silently, whereas the old code had explicit `{ value: string } | undefined` casts. Optional annotation to keep some safety:

```ts
const settings = Object.fromEntries(
  rows.map((r) => [r.key, r.value]),
) as Record<string, string | undefined>;
```

apps/server/src/routes/models.ts
```diff
@@ -45,20 +45,23 @@ const UNSUITABLE_CLEANUP_MODEL_PATTERN =
...
+  const baseUrl = settings.local_llm_url
+    .replace(/\/+$/, "")
+    .replace(/\/v1$/, "");
 
   const res = await fetch(`${baseUrl}/v1/models`, {
     headers: {
-      ...(keyRow?.value ? { Authorization: `Bearer ${keyRow.value}` } : {}),
+      ...(settings.local_llm_api_key
+        ? { Authorization: `Bearer ${settings.local_llm_api_key}` }
+        : {}),
```
MathurAditya724, line 64: Out of scope for this PR (pre-existing), but while you're in this function: this literal `3000` duplicates the existing `REGISTRY_FETCH_TIMEOUT_MS = 3000` constant. Could optionally fold in the constant.

MathurAditya724 (commented):
Reviewed the consolidation of the two LLM settings queries into a single `WHERE key IN (...)` query. The refactor is behaviorally equivalent to the original and I found no bugs. Verified: `!settings.local_llm_url` covers both a missing row and an empty-string value (matches old `!urlRow?.value`), the API-key `Authorization` gate is unchanged, `baseUrl` normalization is unchanged, and `settings` keys are unique so `Object.fromEntries` can't collide. Two optional, non-blocking notes inline. Safe to merge.

PR #397, opened 2026-07-05
Fix High/Med cleanup prompts to be more consistent

tigerisaac:
This PR addresses the problems within the high and medium cleanup prompts, changing them to be more consistent and work better, especially when enumerating lists.

Benchmarks were run, and all of them were passed. Prompt engineering was basically just done in a loop.
There were no regressions, and this change will hopefully make high and medium modes more consistent.

Here's an example of a sentence it handled correctly:
I've got to get some stuff at the store today, so I want to get one carton of milk, two loaves of bread, three cartons of eggs, and four cups.


MathurAditya724 (approved):
Prompt-text-only change to HIGH/MEDIUM cleanup presets. Single canonical source (CLEANUP_PRESET_PROMPTS) shared by server and renderer, no duplication. Tests assert against the exported constant so they stay green. All CI passing. LGTM.

PR #398, opened 2026-07-05
fix(stt): send vocabulary bias to Freestyle Cloud streaming

tigerisaac:
Vocabulary worked for BYOK Soniox but not for Freestyle Cloud, which also transcribes via Soniox. buildAsrVocabularyBias had no freestyle-cloud case (so the bias was never built), and the cloud provider's start message never forwarded it.

- vocabulary-bias: freestyle-cloud shares the soniox branch, producing a soniox-context bias
- providers/freestyle-cloud: forward the context object as `vocabulary` in the start message (also carried on reset)

Pairs with the freestyle-cloud repo change that receives the field and sets Soniox's session `context`.

### DIFF
--- a/apps/server/src/lib/streaming/providers/freestyle-cloud.ts
+++ b/apps/server/src/lib/streaming/providers/freestyle-cloud.ts
@@ -5,6 +5,7 @@ import {
   freestyleCloudStreamWsUrl,
   transcribeWithFreestyleCloud,
 } from "../../freestyle-cloud.js";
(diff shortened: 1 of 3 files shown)


MathurAditya724 (approved):
Verified: fallthrough case produces a soniox-context bias for freestyle-cloud, forwarded on start and reset via the shared sonioxContextFromBias helper (DRY). Bias is wired end-to-end through stream.ts. No runtime circular import (type-only back-edge). Added field is backward-compatible. Tests and all CI green. LGTM.

PR #407, opened 2026-07-06
feat: add "Always on" option to MLX model warming (never unload)

Maheidem:
### What

Adds an **"Always on"** position past the 10-minute maximum on the MLX model-warming slider, so the local `mlx-asr` model can stay resident until the app quits instead of unloading after at most 10 minutes.

### Why

For heavy dictation users, the model unloading after ≤10 min idle means a cold reload (~2–3s) on the next dictation. An always-resident option removes that at the cost of keeping the model in memory.

### How

Uses a sentinel keep-alive value of `-1` ("never unload"):

- **`server.ts`** — `getMlxAsrKeepAliveMinutes()` passes `-1` through instead of clamping to `0`; `scheduleUnload()` returns early without arming the unload timer when the value is `-1`.
- **UI** (`constants.ts` / `utils.ts` / `mlx-memory-section.tsx`) — the slider gains one step past `MAX` for "Always on", the clamp preserves the `-1` sentinel, and the description explains the behavior.

On app quit the worker is still stopped (existing `process.exit` handler), so "Always on" only affects idle retention, not shutdown.

### Validation

Built on macOS arm64 (`build:mac`), installed and run. Selecting "Always on" persists `-1`; the worker stayed resident for 36 minutes across a 12-minute idle gap with no unload (vs. the old cap where it would unload at 10 min). Finite values (0–10) behave exactly as before.

### DIFF
--- a/apps/electron/src/renderer/src/pages/models/constants.ts
+++ b/apps/electron/src/renderer/src/pages/models/constants.ts
@@ -1,2 +1,4 @@
 export const DEFAULT_MLX_KEEP_ALIVE_MINUTES = 10;
 export const MAX_MLX_KEEP_ALIVE_MINUTES = 10;
+// Sentinel meaning "always on" (never unload the model). Stored as -1.
+export const MLX_KEEP_ALIVE_ALWAYS = -1;
--- a/apps/electron/src/renderer/src/pages/models/mlx-memory-section.tsx
+++ b/apps/electron/src/renderer/src/pages/models/mlx-memory-section.tsx
@@ -9,9 +9,12 @@ import {
 import { Slider } from "@renderer/components/ui/slider";
 import { Cpu } from "lucide-react";
 
-import { MAX_MLX_KEEP_ALIVE_MINUTES } from "./constants";
+import { MAX_MLX_KEEP_ALIVE_MINUTES, MLX_KEEP_ALIVE_ALWAYS } from "./constants";
 
 function mlxKeepAliveDescription(minutes: number): string {
+  if (minutes === MLX_KEEP_ALIVE_ALWAYS) {
+    return "Keep the model loaded in memory at all times, until you quit Freestyle. Fastest repeat use, uses the most RAM.";
+  }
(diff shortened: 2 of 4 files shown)


tigerisaac (approved):
Looks good & clean to ship.

MathurAditya724 (commented):
Reviewed the full diff and traced the round-trip. This is a clean, well-scoped change and the core logic is correct:

- The `-1` sentinel guard in `scheduleUnload()` is placed *before* the `delayMs <= 0` branch, which is exactly right — without it, `-1 * 60_000` would have fallen into the immediate-unload path. Good catch.
- `startMlxInBackground()` gates on `=== 0`, so `-1` still prewarms — correct for "always on".
- The save path round-trips: `clampMlxKeepAliveMinutes(-1)` preserves the sentinel, `saveMlxKeepAliveMinutes` skips the `next !== 0` immediate-stop, and the settings PUT has no value schema on `mlx_asr_keep_alive_minutes` so `-1` persists and `applyMlxAsrRetentionPolicy()` applies it live.

CI is green (all three platform builds, server/E2E tests, typecheck, knip, lint). I also ran the server suite (218 passing) and electron typecheck locally — clean.

Two minor, non-blocking notes:

**Duplicated sentinel constant.** `MLX_KEEP_ALIVE_ALWAYS` (renderer `constants.ts`) and `KEEP_ALIVE_ALWAYS` (server `mlx-asr/server.ts`) are independent `-1` definitions. Same story as the pre-existing `MAX_MLX_KEEP_ALIVE_MINUTES` / `MAX_KEEP_ALIVE_MINUTES` pair, so this follows the current convention — but if these ever drift the UI and server disagree silently. Not worth blocking; flagging for awareness.

**Slider endpoint label.** The right-hand caption changed from "Keep warm 10 min" to "Always on". That's accurate for the new last step, but the 10-min position is now second-to-last with no caption. Minor; the description text below covers it.

No behavior concerns. LGTM once a maintainer signs off.

PR #408, opened 2026-07-06
Feat/tone page rework

tigerisaac:
Adds off button and turns all tones off by default. 
Additionally hardens our medium prompt and makes it default

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -1253,24 +1253,30 @@ function resetOnboarding(): void {
   showSettingsWindow("/onboarding");
 }
 
-// Dev-only: simulate a new user's locked Tone page (setup hero).
-// Sets cleanup_tone_enabled to false explicitly so grandfather auto-opt-in
-// (undefined + llm_cleanup on) does not immediately unlock the page.
+// Dev-only: reset every sector tone to off (fresh onboarding state).
 async function resetToneSetup(): Promise<void> {
+  const toneKeys = [
+    SETTINGS_KEYS.cleanupPersonalTone,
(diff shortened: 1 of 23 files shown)


MathurAditya724 (commented):
Reviewed the full diff. Core logic is solid — the `off` sector-tone plumbing, the `getEffectiveCleanupTones()` consolidation, and gating app-context capture behind `needsAppContextForCleanup()` are all consistent across the batch, cloud, and streaming paths. All 219 server tests pass and electron typecheck is clean. Two minor things:

**Dead code — `wide` prop on `TextMessagePreview`**
`apps/electron/src/renderer/src/components/tone-previews/text-message-preview.tsx` adds a `wide` prop (type + default + two internal `cn()` branches), but the only call site (tone.tsx:989) never passes it, so `wide` is always `false`. Either wire it up where the cramped-container layout is needed or drop it to keep the diff clean.

**Test gap — the "all tones off" gating path**
`areAllCleanupTonesOff` / `resolveAppContextForCleanup` / `needsAppContextForCleanup` are the lynchpin of "skip app-context capture when all sectors are off" — both a behavior and a privacy guarantee (frontmost app isn't captured/sent). Right now they're only exercised via a mock passthrough in `history-pause-transcribe.test.ts`. Worth a direct test asserting `resolveAppContextForCleanup(ctx)` returns `null` when all tones are off (and when cleanup is disabled) and passes `ctx` through otherwise. `prompts.test.ts` already covers the prompt-block side well.

Nothing blocking. The prompt hardening looks careful — the retraction-vs-clarification distinction and the symbol/date/time/money reconstruction rules are well exemplified, and the MEDIUM-default + MUST NOT normalize note stay consistent.

PR #414, opened 2026-07-07
Users of an old Qwen3-ASR version must update to the newest one upon use.

matteo8p:
# Overview 

In this PR, we make a change such that when you use Qwen3-ASR on the broken version of MLX, it automatically updates to the newest version.

### DIFF
--- a/apps/electron/src/renderer/src/pages/tone.tsx
+++ b/apps/electron/src/renderer/src/pages/tone.tsx
@@ -777,6 +777,7 @@ function CleanupTonePanel({
           {CLEANUP_OPTIONS.map((option, index) => {
             const selected = option.value === value;
             return (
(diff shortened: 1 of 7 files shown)


apps/electron/src/renderer/src/pages/tone.tsx
```diff
@@ -1082,6 +1083,7 @@ function SubsetTonePanel<T extends string>({
           {options.map((option, index) => {
             const selected = option.value === value;
             return (
+              // biome-ignore lint/a11y/useSemanticElements: roving-tabindex radiogroup on styled buttons; <input type="radio"> would need a full restyle of the card layout.
```
matteo8p, line 1086: These extra non-mlx related things are because of `pnpm format` fixes

PR #422, opened 2026-07-07
Remove extra bottom spacing on the settings page 

### LINKED ISSUE #417: fix: settings page bottom padding
Don't like this blank spacing at the bottom in the settings page, can we remove that so that its flushed with the borders like the other pages?

Image

GatikTulsiani07:
## What changed

This PR removes the extra blank space at the bottom of the Settings page.

The root cause was that the Settings page uses the shared `responsive-page-scroll` class on a grid wrapper rather than the actual scroll container. Since that shared class includes bottom padding, it created a fixed empty space below the page content.

Instead of changing the shared CSS, this PR overrides the bottom padding only for the Settings page so the layout matches the other pages while keeping the existing scrolling behavior unchanged.

## Where I checked

1) Settings page
2) History page
3) Models page

## Validation

Ran:

1) `pnpm biome check`
2) `pnpm --filter @freestyle-voice/electron typecheck:web`

Both passed.

## Screenshot 
Screenshot 2026-07-07 at 2 15 48 PM

Closes #417

### DIFF
--- a/apps/electron/src/renderer/src/pages/settings.tsx
+++ b/apps/electron/src/renderer/src/pages/settings.tsx
@@ -1312,7 +1312,7 @@ function NetworkPanel(): React.JSX.Element {
           )}
         />
       </Row>
-      <div className="border-border bg-secondary/40 text-muted-foreground mt-1 flex items-start gap-2.5 rounded-[10px] border px-3.5 py-3 text-[12px] leading-[1.55]">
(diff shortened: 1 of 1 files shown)


MathurAditya724:
Can we add a bit off spacing in the bottom of this tab in settings, just below the "The app also..." card, then we can merge it
Screenshot 2026-07-07 at 6 48 08 PM

GatikTulsiani07:
Updated this with the small bottom spacing below the final Network tab info card.

Kept the change scoped to `apps/electron/src/renderer/src/pages/settings.tsx`.

Validation:
- `pnpm biome check`
- `pnpm --filter @freestyle-voice/electron typecheck:web`
Screenshot 2026-07-07 at 9 25 22 PM

Both passed.

PR #424, opened 2026-07-07
feat: auto-delete transcription history after a configurable period

### LINKED ISSUE #421: fix: clear history after a certain amount of time.
Right now we can only pause history or have to manually delete the transcription history. In place of this, it would be nice if we can have a scheduling system. So basically a cron job which automatically deletes history after a certain period of time. This shouldn't be running like every minute. This could be something which runs once a day. So we could have options of clearing history in 7 days, 30 days, custom days or none, and the idea being that every day a cron runs and deletes the old messages.

Image

dchaudhari7177:
## Summary

Closes #421

Adds a scheduling system that automatically deletes old transcription history, so users no longer have to choose between pausing history and manually clearing it.

## How it works

- **New setting** `history_retention_days` — configured from **Settings → Data → Auto-delete history** with the options suggested in the issue: **Never / After 7 days / After 30 days / Custom** (custom takes a number of days).
- **Daily sweep** — the server runs a purge once every 24 hours (not every minute, per the issue), plus once at server start so a restart never postpones an overdue cleanup. The timer is `unref()`'d so it can't keep the process alive during shutdown.
- **Immediate effect** — writing the setting also purges expired entries right away (mirroring the existing `applyMlxAsrRetentionPolicy` pattern), so picking "7 days" cleans up old sessions without waiting for the next sweep.
- **Validation** — value must be a whole number of days between 1 and 3650; empty string disables auto-deletion. Enforced server-side in `PUT /api/settings/:key` via a new schema in `@freestyle-voice/validations`, consistent with the other constrained settings keys.

## Tests

- 5 new tests in `apps/server/tests/api.test.ts` covering: no-op when unconfigured, purging only entries older than the window, immediate purge on setting write, empty-value disable, and rejection of malformed values (`abc`, `0`, `-5`, `1.5`, `7 days`, out-of-range).

## Checks

- `vitest run tests/api.test.ts` — 53/53 pass (the 3 `mlx-runtime` failures in the full suite are pre-existing on `main` in my Windows environment and unrelated)
- `tsc --noEmit` (server) and `pnpm --filter @freestyle-voice/electron typecheck:web` pass
- Biome ran clean via the pre-commit hook
- New UI strings added to `en.json` and `template.json` (other locales fall back to English)

### DIFF
--- a/packages/validations/src/settings.ts
+++ b/packages/validations/src/settings.ts
@@ -70,6 +70,28 @@ export const proxyUrlSettingSchema = z
 /** Filesystem path to a custom CA certificate bundle. Empty string clears it. */
 export const caCertPathSettingSchema = z.string().max(4096);
 
+export const HISTORY_RETENTION_DAYS_MAX = 3650;
+
+export function parseRetentionDays(
+  value: string | null | undefined,
+): number | null {
+  if (value == null) return null;
+  const trimmed = value.trim();
+  if (!/^\d+$/.test(trimmed)) return null;
+  const days = Number(trimmed);
+  if (days < 1 || days > HISTORY_RETENTION_DAYS_MAX) return null;
+  return days;
+}
+
+export const historyRetentionDaysSettingSchema = z
+  .string()
+  .refine(
+    (value) => value.trim() === "" || parseRetentionDays(value) !== null,
+    {
+      message: `Retention must be a whole number of days between 1 and ${HISTORY_RETENTION_DAYS_MAX} (or empty to disable)`,
+    },
+  );
+
 /**
  * Combined shape for the Network settings form. The renderer drives a
  * react-hook-form with this schema so its inline validation matches exactly
--- a/apps/electron/src/renderer/src/locales/en.json
+++ b/apps/electron/src/renderer/src/locales/en.json
@@ -187,6 +187,13 @@
     "data": {
       "pauseHistory": "Pause history",
       "pauseHistoryDesc": "Stop saving new transcription sessions to history until you turn this off.",
+      "autoDelete": "Auto-delete history",
+      "autoDeleteDesc": "Automatically delete saved sessions older than the selected period. Runs once a day.",
+      "autoDeleteNever": "Never",
+      "autoDelete7": "After 7 days",
+      "autoDelete30": "After 30 days",
+      "autoDeleteCustom": "Custom",
+      "autoDeleteDays": "days",
       "history": "Transcription history",
       "historyDesc": "Permanently delete every saved session — this can't be undone.",
       "clearHistory": "Clear history",
--- a/apps/electron/src/renderer/src/locales/template.json
+++ b/apps/electron/src/renderer/src/locales/template.json
@@ -188,6 +188,13 @@
     "data": {
       "pauseHistory": "Pause history",
       "pauseHistoryDesc": "Stop saving new transcription sessions to history until you turn this off.",
+      "autoDelete": "Auto-delete history",
+      "autoDeleteDesc": "Automatically delete saved sessions older than the selected period. Runs once a day.",
+      "autoDeleteNever": "Never",
+      "autoDelete7": "After 7 days",
+      "autoDelete30": "After 30 days",
+      "autoDeleteCustom": "Custom",
+      "autoDeleteDays": "days",
       "history": "Transcription history",
       "historyDesc": "Permanently delete every saved session — this can't be undone.",
       "clearHistory": "Clear history",
--- a/apps/electron/src/renderer/src/pages/settings.tsx
+++ b/apps/electron/src/renderer/src/pages/settings.tsx
@@ -1,6 +1,8 @@
 import {
+  HISTORY_RETENTION_DAYS_MAX,
   type NetworkSettingsForm,
   networkSettingsFormSchema,
+  parseRetentionDays,
 } from "@freestyle-voice/validations";
 import { zodResolver } from "@hookform/resolvers/zod";
 import { KeyComboDisplay } from "@renderer/components/key-combo";
@@ -122,6 +124,10 @@ export default function SettingsPage(): React.JSX.Element {
   const [pillPosition, setPillPosition] = useState("bottom-center");
   const [soundEnabled, setSoundEnabled] = useState(true);
   const [historyPaused, setHistoryPaused] = useState(false);
+  const [historyRetention, setHistoryRetention] = useState<
+    "never" | "7" | "30" | "custom"
+  >("never");
+  const [customRetentionDays, setCustomRetentionDays] = useState("90");
   const [audioPlaybackMode, setAudioPlaybackMode] =
     useState<AudioPlaybackMode>("off");
   const [updateAvailable, setUpdateAvailable] = useState<string | null>(null);
@@ -160,6 +166,16 @@ export default function SettingsPage(): React.JSX.Element {
         label:
           t(`settings.recording.transcriptionLanguages.${l.id}`) || l.label,
       })),
+    ],
+    [t],
+  );
+
+  const retentionOptions = useMemo(
+    () => [
+      { value: "never", label: t("settings.data.autoDeleteNever") },
+      { value: "7", label: t("settings.data.autoDelete7") },
+      { value: "30", label: t("settings.data.autoDelete30") },
+      { value: "custom", label: t("settings.data.autoDeleteCustom") },
     ],
     [t],
   );
@@ -310,6 +326,18 @@ export default function SettingsPage(): React.JSX.Element {
     if (s[SETTINGS_KEYS.outputMode]) setOutputMode(s[SETTINGS_KEYS.outputMode]);
     if (s[SETTINGS_KEYS.soundEnabled] === "false") setSoundEnabled(false);
     if (s[SETTINGS_KEYS.historyPaused] === "true") setHistoryPaused(true);
+
+    const retentionDays = parseRetentionDays(
+      s[SETTINGS_KEYS.historyRetentionDays],
+    );
+    if (retentionDays !== null) {
+      if (retentionDays === 7 || retentionDays === 30) {
+        setHistoryRetention(String(retentionDays) as "7" | "30");
+      } else {
+        setHistoryRetention("custom");
+        setCustomRetentionDays(String(retentionDays));
+      }
+    }
 
     // Audio playback mode with legacy fallback chain (new key → paused → duck).
     if (s.audio_playback_mode) {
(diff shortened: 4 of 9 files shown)


MathurAditya724 (commented):
Reviewed the full diff against #421 and traced the paths. This is a well-built feature — the daily `setInterval` with an immediate sweep + `unref()`, the immediate purge on setting write (mirroring `applyMlxAsrRetentionPolicy`), server-side schema validation, and the parameterized `datetime('now', ?)` delete are all done right. I confirmed `created_at` uses the schema default `datetime('now')` (UTC), so the string comparison in the purge is sound. Ran the server suite locally (217 pass) and electron typecheck is clean. Your note about the 3 mlx-runtime failures being Windows-only checks out — they pass on Linux. CI is green.

Test coverage is genuinely good — the no-op/window/immediate/disable/malformed cases are all covered.

One real bug and two nits:

**Bug — out-of-range custom value fails silently.** In `settings.tsx`, `handleCustomRetentionDaysChange` accepts up to 4 digits (`slice(0, 4)` → max 9999) and saves any value `>= 1`, but the schema caps at `HISTORY_RETENTION_DAYS_MAX` (3650). Typing e.g. `9999` sends a PUT that the server rejects with 400, which is swallowed by `.catch(() => {})`. The select still shows \"Custom / 9999\" but nothing is persisted, and there's no error surfaced — the user thinks retention is set when it isn't, and it reverts on next load. Clamp the input to `HISTORY_RETENTION_DAYS_MAX` (import the constant from `@freestyle-voice/validations` — it's already exported) or surface the validation error. This is the one thing worth fixing before merge.

**Nit — duplicated validation logic.** `getHistoryRetentionDays()` in `history-store.ts` re-implements the digit/range check that `historyRetentionDaysSettingSchema` already encodes. Since writes are validated at the route, the store-side parse is a second source of truth that could drift. Not blocking, but reusing the schema (or a shared parse helper) would keep them in lockstep.

**Nit — unused export.** `stopHistoryRetentionSweep` is exported and documented as \"used by tests and shutdown paths\" but has no callers in either. Either wire it into the tests (a `stopHistoryRetentionSweep()` in an `afterAll` would be tidy since the interval is created at import) or drop it.

Nice work overall — fix the silent-failure case and this is in good shape.

MathurAditya724 (changes requested):
Went back through the UI side of this more carefully and there's one thing I'd want fixed before this goes in, plus a few consistency cleanups. Solid feature overall though — the server logic is genuinely well done.

## Needs fixing

**Custom days over 3650 fails silently.** The input takes 4 digits (`slice(0, 4)`, so up to 9999) but the schema caps at 3650. Type anything in 3651–9999 and the PUT comes back 400, which the `.catch(() => {})` just eats. The dropdown still shows "Custom / 9999" so it *looks* saved, but nothing persisted and it snaps back on the next load. Easiest fix is to clamp the input to `HISTORY_RETENTION_DAYS_MAX` (it's already exported from `@freestyle-voice/validations`) — then there's no error to surface at all.

**This is really a symptom of the validation living in three places.** Right now the "digits, >= 1, <= 3650" rule exists in:
- `historyRetentionDaysSettingSchema` — has the upper bound
- `getHistoryRetentionDays()` in the store — no upper bound
- the renderer (load effect + the two save handlers) — no upper bound

They already disagree, which is exactly what lets the bug through. I'd pull a single `parseRetentionDays(value): number | null` into the validations package and call it from all three. Kills the bug and the duplication in one go.

## Nice to have (not blocking)

- **`settings.tsx` — hardcoded `SelectItem`s.** Every other Select in this file maps over an options array (`microphoneOptions`, `languageOptions`). Mind pulling `never/7/30/custom` into a `RETENTION_OPTIONS` const to match?
- **`SelectTrigger` is missing an `id`.** The microphone and language selects both set one (`settings-microphone`, `settings-language`); this one doesn't. Small consistency thing.
- **Add `min-w-0` to the flex row** (or let it wrap). The select + input + "days" span are all in a fixed-width flex row; on a narrow pane the label can't shrink. Minor since the widths are fixed, but it's the pattern the guidelines nudge toward.

The `aria-label` on the custom input and the daily-sweep/immediate-purge logic are all good. Fix the silent-failure case (ideally via the shared parser) and I'm happy.

packages/validations/src/settings.ts
```diff
@@ -70,6 +70,26 @@ export const proxyUrlSettingSchema = z
 /** Filesystem path to a custom CA certificate bundle. Empty string clears it. */
 export const caCertPathSettingSchema = z.string().max(4096);
 
+/** Upper bound on the history auto-delete window (10 years). */
+export const HISTORY_RETENTION_DAYS_MAX = 3650;
+
+/**
```
matteo8p: @dchaudhari7177 Hey Dipak, can you remove code comments from this PR? Thank you.

matteo8p (changes requested):

dchaudhari7177:
Thanks for the thorough reviews @MathurAditya724 @matteo8p — all addressed in the latest push (rebased onto `main`).

**Silent-failure bug (custom > 3650).** Fixed at the root by consolidating the "digits, 1–3650" rule into a single `parseRetentionDays(value): number | null` in `@freestyle-voice/validations`. It's now the one source of truth for all three call sites:
- `historyRetentionDaysSettingSchema` — `refine` is now `value.trim() === "" || parseRetentionDays(value) !== null`
- `getHistoryRetentionDays()` in the store — just `return parseRetentionDays(readSetting(...))`
- the renderer — load effect and both save handlers

The custom input also clamps to `HISTORY_RETENTION_DAYS_MAX`, so a value like `9999` becomes `3650` on entry — there's no longer a rejected PUT to swallow, and the three places can't drift.

**`stopHistoryRetentionSweep` had no caller.** Wired it in: an `afterEach(() => stopHistoryRetentionSweep())` in the retention suite, plus a new test that calls `startHistoryRetentionSweep()` twice and asserts the immediate purge ran and the second call is a no-op (idempotency). So both the sweep start and stop are now exercised.

**Consistency nits (all done):**
- `settings.tsx` now maps a `retentionOptions` array (matching `microphoneOptions`/`languageOptions`) instead of hardcoded `SelectItem`s.
- `SelectTrigger` gets `id="settings-history-retention"`.
- `min-w-0` added to the flex row.

**Comments** — removed the added code comments per @matteo8p.

Server suite passes (the only failures are the 3 Windows-only `mlx-runtime` ones we discussed; they pass on Linux/CI), electron typecheck + tests are clean, and biome ran via the pre-commit hook. Let me know if anything else stands out.

MathurAditya724:
Can you resolve the merge issue?

dchaudhari7177:
Rebased onto latest `main` (0.4.4) and resolved the conflict. The clash was in `settings.tsx`: `main` migrated settings loading to the batch `settingsQuery`, so I moved the retention seeding into that seed effect (reading `s[SETTINGS_KEYS.historyRetentionDays]` via `parseRetentionDays`) and dropped the old per-key `$get` block. No behavior change from the review revision. Squashed to a single commit; typecheck, `api.test.ts` (54), and the electron tests are green locally. Ready for another look.

MathurAditya724 (commented):
Reviewed the rebase specifically against the settings-batch work that just landed in `main` via #436 (squash `29d6929`). **No changes needed — the rebase is correctly aligned with every pattern #436 introduced.**

Confirmed:

1. **Seed-once effect** — the retention seeding (`parseRetentionDays(s[SETTINGS_KEYS.historyRetentionDays])`) sits *inside* the shared `settingsSeeded`-guarded seed effect (right after the `historyPaused` seed), so it reads the persisted value once and won't re-seed/clobber in-progress edits on a later invalidation. Correct placement.

2. **Save-handler convention** — `saveHistoryRetention` / `handleHistoryRetentionChange` mirror the sibling handlers exactly: update local `useState` → `$put` → `.catch(() => {})`. Correctly does **not** patch the `["settings-all"]` cache — the in-place `setQueryData` patch from #436 is specific to `NetworkPanel` (a separate child component with its own seed guard). Inside `SettingsPage`, the seed-once guard means a stale cache never re-seeds, so no patch is needed. Matches `handleHistoryPausedToggle` / `handleAudioPlaybackModeChange`, the handlers this code sits next to.

3. **Server-side validation** — schema in `@freestyle-voice/validations` enforced through `PUT /api/settings/:key` is consistent with the validation hardening from #436.

One minor UX note (not a blocker, unrelated to the rebase): in `handleHistoryRetentionChange`, the `"custom"` branch only persists when `customRetentionDays` already parses valid. If a user selects "Custom" while the field is empty/invalid, nothing is written until they type a valid number — the previous value stays in effect server-side. Probably intended, just flagging.

LGTM on the rebase.

dchaudhari7177:
Thanks for re-reviewing the rebase against #436 so carefully @MathurAditya724 — glad the seed-once placement and the no-cache-patch handler convention line up.

On the custom-field UX note: it's intended. The guard deliberately avoids writing an empty/invalid custom value so it can't clobber the server-side retention with garbage. In the normal path `customRetentionDays` defaults to `"90"`, so selecting **Custom** persists right away; the "nothing written" case only happens if the field has been cleared to empty, and it self-corrects the moment a valid number is typed (and on reload the seed effect falls back to the server's value). Happy to add a small fallback-to-last-valid if you'd prefer, but I'd lean toward leaving it as-is to avoid persisting an unintended value.

Also — the code comments @matteo8p flagged were removed in the review revision, so that one should be resolved too whenever you get a chance to take another look.

MathurAditya724 (approved):
Approving from my side. I traced the feature end-to-end and it's correct at every layer:

- **Purge SQL** — `DELETE ... WHERE created_at < datetime('now', '-N days')` compares against `created_at TEXT DEFAULT (datetime('now'))`, so both sides are identical-format UTC strings; the comparison is chronologically correct.
- **Sweep timer** — single interval, double-start guarded, `unref()`'d, immediate purge on start so a restart never postpones an overdue cleanup, and `try/catch → captureException` so a failed purge can't kill the timer.
- **Write path** — validation slots into the existing per-key `else if` chain (matches `caCertPathSettingSchema`), and the immediate `purgeExpiredHistory()` fires after the INSERT so it reads the freshly-written value.
- **Startup** — sweep starts after `initServerPlugins()` and `getDb()` is lazy-init, so no "DB not ready" risk.
- **`parseRetentionDays`** — strict (`/^\d+$/`, range 1–3650, empty = disable) and the single source of truth for schema + store + renderer, so the three can't drift.
- **Renderer** — seeded once inside the shared `settingsSeeded` effect (won't clobber edits); custom input clamps to `HISTORY_RETENTION_DAYS_MAX` on entry.

The earlier findings (silent-failure on custom > 3650, unwired `stopHistoryRetentionSweep`, code comments) are all addressed and verified in the current diff, and the rebase lines up cleanly with the settings-batch work from #436. CI is fully green.

The custom-empty-field behavior is intentional and reasonable as you described — leaving as-is.

Deferring the final merge to @matteo8p for their pass on the comment cleanup they flagged.

matteo8p (approved):
Looks good, let's ship it.

PR #425, opened 2026-07-07
feat: diff mode for AI-edited history entries

### LINKED ISSUE #420: feat: diff mode in history
Currently it's hard to figure out if the issue was in transcription or post processing, because the resulting history is the final output by the service. The idea is to store both the outputs of transcription and post processing, and by toggling some filters -

1. for seeing both the things simultaneously and,
2. seeing a diff between both the outputs.

This would be a nice addition for debugging results

dchaudhari7177:
## Summary

Closes #420

History entries only show the final output, so when a result looks wrong there's no way to tell whether the transcription or the post-processing was at fault. Both outputs are already stored (`raw_text` / `cleaned_text`) — this adds a way to see them together.

## What it does

Each AI-edited entry gets a new **diff toggle** (file-diff icon, next to the existing undo/redo AI-edit button). Toggling it renders both outputs in a single reading pass:

- words the post-processing **removed** are struck through (destructive tint)
- words it **added** are highlighted (primary tint)
- unchanged words render plainly

This covers both asks in the issue — seeing the two outputs simultaneously and seeing the diff between them. The existing undo/redo toggle is disabled while the diff is shown (the diff already displays both versions).

## Implementation notes

- Word-level LCS diff (`diffWords`), ~60 lines, no new dependency.
- Computed only when the user toggles diff mode and memoized per entry. A worst-case 2000-word transcript diffs in ~170 ms; typical dictations are sub-millisecond.
- Beyond a 2000×2000-word cap it degrades gracefully to "full raw struck through + full cleaned highlighted", which still shows both outputs.
- Verified segment reconstruction round-trips exactly: same+del segments rebuild the raw text, same+add segments rebuild the cleaned text.
- Entries without an AI edit are unchanged (no diff button).

## Checks

- `pnpm --filter @freestyle-voice/electron typecheck:web` passes
- Biome ran clean via the pre-commit hook

### DIFF
--- a/apps/electron/package.json
+++ b/apps/electron/package.json
@@ -25,6 +25,7 @@
     "build:linux": "npm run compile:native && electron-vite build && electron-builder --linux --publish never",
     "download:whisper-cpp": "node scripts/download-whisper-cpp.mjs",
     "download:whisper-cpp:all": "node scripts/download-whisper-cpp.mjs --all",
+    "test": "vitest run",
     "test:e2e": "playwright test"
   },
   "dependencies": {
@@ -40,6 +41,7 @@
     "@tanstack/react-query": "^5.101.2",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
+    "diff": "^8.0.4",
     "electron-updater": "^6.3.9",
     "freestyle-voice": "workspace:*",
     "hono": "^4.12.22",
@@ -78,6 +80,7 @@
     "react-dom": "^19.2.1",
     "tailwindcss": "^4.3.0",
     "typescript": "^5.9.3",
-    "vite": "^7.2.6"
+    "vite": "^7.2.6",
+    "vitest": "^4.1.7"
   }
 }
--- a/apps/electron/src/renderer/src/lib/history-diff.test.ts
+++ b/apps/electron/src/renderer/src/lib/history-diff.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from "vitest";
+import { diffWords } from "./history-diff";
+
+/** Reconstruct the raw text from the segments the diff kept from it. */
+function rebuildRaw(segments: ReturnType<typeof diffWords>): string {
+  return segments
+    .filter((s) => s.type !== "add")
+    .map((s) => s.text)
+    .join("");
+}
+
+/** Reconstruct the cleaned text from the segments the diff kept from it. */
+function rebuildCleaned(segments: ReturnType<typeof diffWords>): string {
+  return segments
+    .filter((s) => s.type !== "del")
+    .map((s) => s.text)
+    .join("");
+}
+
+describe("diffWords", () => {
+  it("round-trips: del+same rebuild raw, add+same rebuild cleaned", () => {
+    const raw = "um so I think we should uh ship it tomorrow";
+    const cleaned = "I think we should ship it tomorrow.";
+
+    const segments = diffWords(raw, cleaned);
+
+    expect(rebuildRaw(segments)).toBe(raw);
+    expect(rebuildCleaned(segments)).toBe(cleaned);
+  });
+
+  it("marks removed and added words", () => {
+    const segments = diffWords("the quick brown fox", "the quick red fox");
+
+    const removed = segments
+      .filter((s) => s.type === "del")
+      .map((s) => s.text.trim());
+    const added = segments
+      .filter((s) => s.type === "add")
+      .map((s) => s.text.trim());
+
+    expect(removed).toContain("brown");
+    expect(added).toContain("red");
+  });
+
+  it("returns a single unchanged segment when texts are identical", () => {
+    const segments = diffWords("hello world", "hello world");
+
+    expect(segments).toEqual([{ type: "same", text: "hello world" }]);
+  });
+
+  it("diffs spaceless scripts word-by-word, not as one blob (CJK)", () => {
+    // Japanese: only the trailing です is removed.
+    const segments = diffWords(
+      "昨日私は東京に行きましたです",
+      "昨日私は東京に行きました",
+    );
+
+    // There must be an unchanged portion — the whole thing is not one del+add pair.
+    expect(segments.some((s) => s.type === "same")).toBe(true);
+    expect(
+      segments
+        .filter((s) => s.type === "del")
+        .map((s) => s.text)
+        .join(""),
+    ).toBe("です");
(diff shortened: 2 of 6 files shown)


MathurAditya724 (changes requested):
Really like this feature — being able to see raw vs cleaned inline is exactly what #420 was after, and the rendering side is well done (semantic `<del>`/`<ins>`, `dir="auto"`, trailing whitespace kept outside the styled span, memoized per toggle). My concern is the diff engine itself, specifically around non-Latin languages, and I think it needs a change before merge.

## The tokenizer breaks for CJK / spaceless scripts

`tokenizeWords` splits on whitespace (`/\S+\s*/g`). That's fine for English, but Chinese, Japanese, Korean, and Thai don't put spaces between words — and all four are first-class transcription languages here (`zh`, `ja`, `ko` are even `onboarding: true` in `languages.ts`). For those, the whole sentence tokenizes as a *single* token, so raw != cleaned means `tokenEquals` fails on the one token and the diff collapses to "entire raw struck through + entire cleaned highlighted." Confirmed it locally:

```
"昨日私は東京に行きましたです"  ->  ["昨日私は東京に行きましたです"]   // 1 token
"我今天去了商店买东西"          ->  ["我今天去了商店买东西"]           // 1 token
```

So for ~4 supported languages the feature effectively shows two colored blobs instead of a diff. The fix doesn't need a dependency: `Intl.Segmenter` (available in Electron 39) does locale-aware word segmentation and handles all of these:

```
new Intl.Segmenter("ja", { granularity: "word" })  ->  ["昨日","私","は","東京","に","行き","ま","した","です"]
new Intl.Segmenter("zh", { granularity: "word" })  ->  ["我","今天","去了","商店","买东西"]
```

The entry doesn't currently carry a language, so passing `undefined` (locale-default segmentation) already beats whitespace splitting; even better if the transcript language is threaded through.

## Do we need custom diff at all?

Honestly the custom LCS is fine *as an approach* — word-level LCS is the right shape and keeping it dependency-free is a real plus. The problem isn't "custom vs library," it's the tokenizer (above) and the cost (below). If we did want a lib, the one that sidesteps tokenization entirely is `diff-match-patch` (char-level Myers, O(ND), script-agnostic) — but it produces sub-word noise on English, so I'd rather keep the word-level LCS + fix the tokenizer with `Intl.Segmenter`. No new dep, works everywhere.

## Performance — the cap is too generous

The PR says "~170 ms worst case," but I benchmarked the actual cap (`DIFF_MAX_TABLE_CELLS = 4M`, ~2000x2000): ~278 ms and a **14.5 MB** `Uint32Array`, allocated synchronously on the render thread when you toggle. That's a quarter-second UI freeze on a long transcript. Two options: drop the cap a lot lower (say 500x500 = ~1MB, still well beyond any real dictation), or switch the traceback to the classic two-row / Hirschberg approach so memory is O(min(n,m)). Given dictations are almost never 2000 words, just lowering the cap is probably enough.

## Minor
- Word-level diffing shows `milk` -> `milk.` as full delete+add rather than an added `.`. Inherent to word granularity and acceptable, just noting it.
- No test for `diffWords`. It's pure and easy to unit test (round-trip: same+del reconstructs raw, same+add reconstructs cleaned; plus a CJK case once the tokenizer is fixed) — would be a good guard.

tl;dr: the feature and UI are great, but the whitespace tokenizer makes it non-functional for CJK/Thai. Swap it for `Intl.Segmenter` (keeps it dependency-free) and tighten the size cap, and I'm on board.

MathurAditya724:
Following up on my review — agreed that nobody wants to maintain a hand-rolled LCS, so let's just use a library. I dug into the options and there's a clean answer that also fixes the CJK problem for free:

**Use `diff` (jsdiff) `diffWords`.** Three reasons it's the right call here:

1. **It's already in the dependency tree.** `diff@8.0.4` is in `pnpm-lock.yaml` as a transitive dep — promoting it to a direct dependency of `apps/electron` costs basically nothing (no new download).
2. **It handles CJK out of the box**, which is the main thing that's broken today. It uses `Intl.Segmenter` internally, so the whitespace-tokenizer issue just goes away. Verified on v8:
   ```
   diffWords("昨日私は東京に行きましたです", "昨日私は東京に行きました")
   ->  [=]昨日私は東京に行きました [-]です
   ```
   And English is clean too — it even handles `milk` -> `milk.` as a `[+].` instead of the full delete+add the custom code produces.
3. BSD-3-Clause, ESM build so `import { diffWords } from "diff"` tree-shakes, ~15M weekly downloads, de-facto standard.

That lets you delete the whole `diffWords` / `tokenizeWords` / `tokenEquals` / `DIFF_MAX_TABLE_CELLS` block (~60 lines) and the perf cap concern along with it — jsdiff uses Myers under the hood, so no 14 MB O(n·m) table. The `DiffText` renderer stays almost as-is; just map jsdiff's `{ value, added, removed }` parts onto your existing `del`/`add`/`same` spans.

So to revise my earlier take: skip `Intl.Segmenter`-by-hand and skip the custom LCS entirely — `diff` gives you both, and it's already here.

dchaudhari7177:
Thanks for the thorough review @MathurAditya724 — agreed on all counts, and your follow-up is the right call. Pushed a2ff4eb which does exactly that: **drops the custom LCS entirely and uses `diff` (jsdiff) `diffWords`.**

- **CJK / spaceless scripts** — fixed. jsdiff segments via `Intl.Segmenter`, so the whitespace-tokenizer problem is gone. The Japanese case you flagged now diffs to `[=]昨日私は東京に行きました [-]です` instead of two colored blobs. There's a unit test asserting exactly this (unchanged portion present, only `です` removed).
- **Performance** — the O(n·m) table and `DIFF_MAX_TABLE_CELLS` cap are deleted along with the custom code; jsdiff is Myers, so no 14 MB allocation on the render thread.
- **`milk` → `milk.`** — now an added `.` rather than full delete+add, as you noted.
- **Dependency** — promoted `diff@8.0.4` (already transitive in the lockfile) to a direct dependency of `apps/electron`; no new download.
- **Test** — the diff logic is extracted to a pure `history-diff` module with unit tests (round-trip reconstruction, add/remove marking, identical-text, and the CJK case). Added a minimal `vitest` setup to the electron package to host them.

The `DiffText` renderer is unchanged. Net diff removes ~90 lines. `typecheck:web`, `biome`, and the new tests all pass locally. Let me know if you'd like the transcript language threaded through to `diffWords` as a follow-up — the entry doesn't currently carry it, so I left segmentation at locale-default for now.

MathurAditya724 (approved):

PR #426, opened 2026-07-07
Replace history filter drawer with the inline filter panel

### LINKED ISSUE #419: fix(ui): filter panel should slide in the UI in the history panel
Currently the filter panel in the history page is a drawer which hides all the transcriptions in the background. It would be nice if this could slide in the UI by shifting the history page a bit, so that both the things are next to each other and visible at the same time.

Also improve this panel so that the filters are instantaneous and we can get rid of the done button. Also if we can improve the UI with more shadcn components that would be nice. Do include the new layouts screenshot in the PR so that its easier to compare

Image

GatikTulsiani07:
## What changed

This PR replaces the History page's modal filter drawer with an inline filter panel rendered inside the History page layout.

The history content now remains visible while filtering instead of being covered by a modal overlay.

Also included:

1) Removed the modal/drawer behaviour.
2) Removed the redundant Done button since filters already apply instantly.
3) Kept existing filter state, date validation, search, pagination, and query behavior unchanged.
4) Added a Filters / Hide Filters toggle.
5) Improved spacing and visual balance for the filter panel.
6) Kept the panel responsive so it remains in-page on narrower layouts.

### Filters closed

WhatsApp Image 2026-07-07 at 15 36 02

### Inline filter panel open
WhatsApp Image 2026-07-07 at 15 36 32

### Narrow layout
Screenshot 2026-07-07 at 3 36 58 PM

## Validation

Ran:

```bash
pnpm biome check

pnpm --filter @freestyle-voice/electron typecheck:web

Both Passed 


closes #419


MathurAditya724:
The UI doesn't look good, can you improve that? I would like the panel to take the full static height. Don't need the narrow UI, it should still open on the side, but we can put the cards in a grid of 2x2 and I feel like we can hide the button "filters" when the panel is open and instead it should be inside the panel

GatikTulsiani07:
ok @MathurAditya724 bro noted i am working on the required things

GatikTulsiani07:
Fixed the date range popover layout as a follow-up.

1) Keeps the two-month shadcn-style range picker.
2) Prevents the popover from covering the History content.
3) Keeps instant filtering and the inline panel behaviour unchanged.

PHOTO-2026-07-08-02-00-51
PHOTO-2026-07-08-02-08-00

MathurAditya724:
2 things, 1. Can you resolve the merge issues? 2. The background of the filter panel is off can you improve that

GatikTulsiani07:
Bro i solved the merge conflicts check the background now 
PHOTO-2026-07-08-11-09-28

matteo8p (approved):
Sweet, thank you so much for putting in this PR

PR #428, opened 2026-07-07
feat: send custom vocabulary to cloud v2 transcribe

MathurAditya724:
## What

Sends the user's **custom vocabulary** to Freestyle Cloud's `/v2/transcribe` so the cloud can bias Whisper recognition toward the speaker's terms.

Companion to **freestyle-voice/cloud#35**, which adds the `vocabulary` field to the cloud v2 route.

## Changes

- `lib/vocabulary.ts`: new `getCloudVocabularyBias()` returning `{ terms?, text? }` (or `undefined`), built from `loadVocabularyTerms()` + the `transcription_prompt` setting. Shape mirrors the cloud contract, so the cloud assembles the recognizer prompt from raw values.
- `lib/freestyle-cloud.ts`: `transcribeWithFreestyleCloud` accepts an optional `vocabulary` and appends it (JSON) to the multipart form. Applied in **both** modes — biasing is independent of cleanup.
- Wired into both cloud call sites: the batch provider (`streaming/providers/freestyle-cloud.ts`, `mode: "raw"`) and the combined path (`routes/transcribe.ts`, `mode: "combined"`).

Dictionary is unchanged — it's applied locally to the cloud response and never sent.

## Compatibility

The `vocabulary` field is optional and ignored by older cloud server versions, so this can ship independently of the cloud PR. Biasing only takes effect once cloud#35 is deployed.

## Verification

- `biome check` clean
- `tsc --noEmit` clean (server)
- `vitest run` — 215 passed
- server build OK

### DIFF
--- a/apps/server/src/lib/freestyle-cloud.ts
+++ b/apps/server/src/lib/freestyle-cloud.ts
@@ -9,6 +9,7 @@ import { createAuthClient } from "better-auth/client";
 import { deviceAuthorizationClient } from "better-auth/client/plugins";
 import type { CloudUser } from "./sessions.js";
 import { CLOUD_TRANSCRIBE_TIMEOUT_MS } from "./streaming/types.js";
+import type { CloudVocabularyBias } from "./vocabulary.js";
 
 export const FREESTYLE_CLOUD_PROVIDER_ID = "freestyle-cloud";
 export const FREESTYLE_CLOUD_TRANSCRIBE_MODEL_ID = "freestyle-cloud/stt";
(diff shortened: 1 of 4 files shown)


matteo8p (changes requested):
I really appreciate you putting in this PR! Looks good to me. Please make sure that vocabulary is also being used in Cloud V2, and we have prompts set up to use this vocabulary.

Do you also think it's worthwhile to send over dictionary terms and have post-processing do the dictionary as well? Right now, we handle dictionary lookups in the Electron app deterministically. This is generally okay, but maybe we also want to support some fuzzy matching. An LLM could be very helpful with that.

matteo8p (approved):

PR #436, opened 2026-07-08
Harden API validation and tighten renderer performance

MathurAditya724:
## Summary

Two related tracks of hardening and performance work across the server, the Electron renderer, the shared validations package, and the profanity-filter plugin UI.

### Backend — request validation & middleware
- Added zod schemas and routed them through `zValidator`, replacing ad-hoc manual body/query parsing: auth device-token, diagnostics (telemetry + client-error), post-process, dictionary import, model config, and history/whisper/mlx query+body.
- New schema files: `packages/validations/src/{auth,diagnostics,post-process}.ts`; extended `dictionary.ts`, `models.ts`, `query.ts`.
- Added `requestId` and a dev-only `logger` middleware (method/path/status via the app logger).
- Added a scoped **30s timeout** on the lightweight CRUD routers; transcription, downloads, and the auth device-poll are intentionally exempt.
- Preserved prior lenient behaviors deliberately (dictionary import still skips blank rows; history dates coerce to `undefined` rather than 400).

### Renderer — React Query & performance
- New `lib/query.ts` with `createQueryClient()` and global defaults (`refetchOnWindowFocus: false`, `staleTime: 1h`, `retry: 1`).
- Migrated `use-models` (polling → `refetchInterval`, memoized derivations), `settings` (~10 GETs → one batch query), `tone`, and `onboarding` status polling to React Query.
- **Route-level code splitting** in `dashboard.tsx` — all non-landing routes are lazy-loaded behind Suspense (confirmed by per-page chunks in the build).
- Replaced the full lucide icon barrel (~1500 icons) with a curated plugin-icon map.
- `React.memo` on dictionary/vocabulary rows; `useCallback`-stable `setPlugins`.

### Plugin — profanity-filter UI
- Memoized the filtered word list (was recomputed every render/keystroke).
- Fixed a case-sensitivity bug in word-name search (raw word vs. lowercased query).
- Memoized `WordRow` with stabilized callbacks.

## Verification
- typecheck (server + electron node/web), biome (clean)
- 210 server tests + 4 electron tests pass
- full `pnpm build` succeeds (all 8 tasks), plugin build clean


MathurAditya724 (commented):
Reviewed the full diff against `main`, ran the checks locally: biome clean, **210 server tests + electron typecheck + server typecheck all pass**. No bugs found — this is solid. A few notes, none blocking.

## Backend — validation & middleware
- The `zValidator` migration follows the pattern already established in `api-keys.ts`/`vocabulary.ts`, so it's consistent with the codebase. Good.
- Default-ordering behavior is preserved: old code defaulted `orderBy` to `-created_at` (DESC); the new schema returns `undefined` when the param is absent and every route falls back to `DESC` in that case. Verified across dictionary/vocabulary/history. ✅
- Lenient behaviors are correctly retained: `dateStringSchema` uses `.catch(undefined)` so malformed history dates are ignored rather than 400'd, and `importDictionarySchema` only asserts the array shape and lets the route skip blank rows. Matches the stated intent.
- One thing to be aware of (not a regression): query params are now strictly validated (`limit` must be `positive().max(200)`, `offset` `min(0)`). The renderer only ever sends valid values (`PAGE_SIZE` / `page*PAGE_SIZE` / `-created_at`), so no practical impact — but any external/manual caller sending `limit=0` now gets a 400 instead of the old silent fallback to 50. Fine, just flagging.
- Timeout scoping looks right — transcription/downloads/device-poll correctly excluded via the prefix allow-list.

## Renderer — React Query & perf
- Lazy/Suspense split verified: all 11 lazy pages have matching `export default`, the single `<Suspense>` wraps `<Routes>` (covers nested `AppShell` routes), and `Today`/`NotFound`/`AppShell` are correctly kept eager.
- Polling migrations are correct: `use-models` `refetchInterval` returns `false` when no download is active (polling stops), and the active→done transition invalidates `["models"]`. Onboarding MLX-probe gating (`!IS_MAC || isFetched`) preserves the old auto-pick timing.
- `React.memo` rows + `useCallback` handlers have correct deps and pass the entry/id as arguments (no stale closure) — memoization is effective and rows still update on data change.
- **Icon map (`helpers.ts`)**: the curated map has ~21 entries vs. the old ~1500-icon barrel. All first-party plugin icons are covered, but any third-party plugin whose manifest `icon` isn't in the list now falls back to the Puzzle icon instead of its chosen one. This matches the stated intent (curated map + puzzle fallback) — just confirm that's the desired contract for plugin authors, or document the supported icon set.
- Minor scope: `pages/today.tsx` extracts the inline query object into a typed `q` var — a functional no-op unrelated to the listed intent. Harmless, just diff noise.

## Plugin — profanity-filter
- Case-sensitivity fix is correct: `e.word.includes(q)` → `e.word.toLowerCase().includes(q)` (`q` is already lowercased). ✅
- `WordRow` memo + stabilized callbacks are correct; `mutation.mutate` identity is stable so the `useCallback` deps are sound.

## deslop
Ran a deslop pass — the diff is clean. The added comments are genuinely explanatory (why the timeout excludes certain routes, why import stays lenient, why `refine` over `.trim()`) and match the repo's existing commenting style. Nothing to strip. Note the `.lore.md` block contains one malformed lore entry (`Switched from plan to build):`), but that's lore-tooling output, not part of this PR's scope.

**Verdict: looks good to merge.** Only actionable item is confirming the intentional icon-fallback contract.

MathurAditya724 (commented):
Reviewed the new commit `d6a0b69` (*Dedupe renderer settings fetches into a single request*). Ran electron typecheck + biome locally — both clean. No bugs.

## What changed
- `app.tsx` pill-startup effect: collapsed up to 5 individual `GET /api/settings/:key` calls (plus 404s for unset keys) into one `GET /api/settings`, reading `soundEnabled` / audio-playback-mode / `outputMode` from the single snapshot.
- New `applyNeedsAppContextForCleanup(settings)` in `cleanup-app-context.ts` — a pure, sync variant that warms the cache from an already-loaded snapshot; `refreshNeedsAppContextForCleanup()` now delegates to it.
- `onboarding.tsx` + `tutorial-demo.tsx`: converted their remaining one-off hotkey fetches to `useQuery(["settings-all"])` so they share the cache other pages already populate.

## Verified correct
- **Audio-playback fallback precedence preserved.** The new `if (mode) / else if (pause...) / else (duck?/off)` chain matches the old sequential-return logic exactly (`audio_playback_mode` → legacy `pause_playback_while_recording` → `audio_ducking_enabled` → `off`).
- **No cleanup-context regression on startup fetch failure.** The startup effect now only warms the cache via `applyNeedsAppContextForCleanup` inside `.then`, so a failed GET leaves `cachedNeedsAppContext === null`. That's safe: the actual gate at record time (`startRecording`, app.tsx:568) still calls `refreshNeedsAppContextForCleanup()`, which re-fetches and falls back to `?? true` on failure — the same safe default as before.
- **`refreshNeedsAppContextForCleanup` is not a dangling import** — still used at app.tsx:568.
- **`["settings-all"]` cache sharing is sound.** The queryFn bodies are byte-identical across settings/tone/onboarding/tutorial-demo, so React Query dedups them onto one cache entry. `tutorial-demo` correctly gates with `enabled: hotkey === undefined`.

## Minor (non-blocking)
- **DRY**: the `["settings-all"]` `queryKey` + `queryFn` pair is now copy-pasted in 4 places (`settings.tsx:294`, `tone.tsx:255`, `onboarding.tsx:205`, `tutorial-demo.tsx:97`). Consider extracting a shared `settingsQueryOptions()` / `useAllSettings()` helper so the fetch shape stays in one spot. `use-models.ts` already names the key via a `QUERY_KEYS.settings` const — reusing that would also prevent a future key-string typo silently splitting the cache.

## deslop
Clean — the added comments are genuinely explanatory (single-request rationale, cache-warming note, JSDoc on the new pure helper) and match this file's existing commenting style. Nothing to strip.

**Verdict: good to merge.** The only follow-up is the optional `["settings-all"]` dedup helper.

MathurAditya724 (commented):
Reviewed `ca6cc2f` (*Extract shared `settingsQueryOptions()` helper*) — this is exactly the dedup I suggested last round, cleanly done. Electron typecheck + biome pass locally.

## Verified
- `settingsQueryOptions()` + `SETTINGS_QUERY_KEY` live in `lib/query.ts` with a clear doc comment; all five consumers (settings, tone, models, onboarding, tutorial-demo) now spread the helper. `use-models` points `MODELS_KEYS.settings` at the shared constant. Fetch shape is in one spot now. ✅
- `tutorial-demo.tsx` correctly dropped its now-unused `getClient` import (confirmed no dangling import).

## One gap I found — and fixed
`tone.tsx` still invalidated with a **raw `["settings-all"]` literal** at the `reload()` callback while everything else moved to the constant — the precise stray-key case `SETTINGS_QUERY_KEY` was introduced to eliminate. It matched today so no live bug, but it defeats the guarantee.

Since it's our PR and a trivial change, I pushed the fix in `dc657fa`: point the invalidate at `SETTINGS_QUERY_KEY`. The only remaining `"settings-all"` string literal in the renderer is now the constant definition itself.

**Verdict: good to merge.**

MathurAditya724 (commented):
Reviewed `364a64c` (*Stop redundant settings refetches on the Models and Network pages*). Electron typecheck + biome clean. No bugs — and this one fixes a real write-amplification bug.

## The actual bug fix (Models page)
Gating the Freestyle-cleanup pairing effect on `m.settingsSeeded` is correct and worth calling out: previously the effect only waited on `m.loading`. Once loading flipped false but *before* the seed effect populated `llmCleanup` from persisted settings, the effect read the initial `useState(false)` → computed `needsSync = true` → re-ran `configureModel` + `setCleanup(true)`, **re-writing the setting on every visit** to the Models page while Freestyle Transcribe was active. Gating on `settingsSeeded` (which specifically tracks the settings→local-state seed, unlike the composite `loading`) closes that. ✅

## Other changes — verified correct
- **`settingsSeeded` ref→state conversion**: needed so `index.tsx` can react to the seed. The seed effect self-guards (`if (!s || settingsSeeded) return`) and adds `settingsSeeded` to its deps — no double-seed, no loop (the guard returns on the re-run triggered by `setSettingsSeeded(true)`). The `keepAlive` seed correctly stays a ref, and the `seededRef` shape change (`{settings,keepAlive}` → `{keepAlive}`) left no dangling `.settings` reference. ✅
- **Skip reload-on-mount (`index.tsx`)**: `prevCloudUserId` seeded with the current id, so the mount pass is a no-op and `reloadModels()` only fires on an actual sign-in/out user change — the queries already self-load, so the old effect was a redundant second fetch. Correct.
- **NetworkPanel cache reuse + patch-on-save**: seeds from the shared `settingsQueryOptions()` cache (drops the two dedicated single-key GETs and the now-removed `loadStringSetting`), and `queryClient.setQueryData(SETTINGS_QUERY_KEY, ...)` patches the cache in place on each save so it stays truthful without a refetch. The `seededRef` seed-once guard means the patched cache won't clobber in-progress edits. Consistent with the seed-once pattern used elsewhere. And it correctly uses the `SETTINGS_QUERY_KEY` constant, not a raw literal. ✅
- No dangling imports: `getClient`/`SETTINGS_KEYS` are still used elsewhere in `settings.tsx` after `loadStringSetting` removal.

## deslop
Clean — comments are explanatory (why the seed gate, why the mount skip, why patch-in-place) and match the file's existing style. `.lore.md` churn is lore-tooling compaction, out of code scope. Nothing to strip.

**Verdict: good to merge.** Nothing to fix here — I didn't push anything.

PR #441, opened 2026-07-08
feat(models): support custom base urls for OpenAI compatible STT APIs

### LINKED ISSUE #415: Support custom base url's for OPENAI-compatible STT API's
## Summary

Freestyle currently supports OpenAI and other cloud Speech-to-Text providers, but there is no way to configure a custom API base URL for OpenAI-compatible STT services.

At the moment, only the local cleanup LLM supports configuring an OpenAI-compatible endpoint through the existing `local_llm_url` setting. The transcription pipeline always uses the SDK defaults (or hardcoded endpoints for realtime), which prevents using self-hosted or compatible OpenAI STT implementations.

Examples include:

1) self-hosted OpenAI-compatible gateways
2) LiteLLM
3) vLLM
4)LM Studio exposing OpenAI-compatible STT APIs
5) other OpenAI-compatible providers

## Current behaviour

For cloud transcription, users can currently configure:

1) STT provider
2) STT model
3) API key

However, the upstream API endpoint is not configurable.

The batch transcription flow relies on the provider SDK defaults, while the OpenAI realtime implementation uses a fixed websocket endpoint.

## Proposed Solution

Introduce optional base URL support for OpenAI-compatible STT providers.

The idea is to:

1)allow an optional custom base URL for supported providers
2) continue using the current default endpoint when no custom URL is configured
3) keep the existing API key and model selection flow unchanged
4) preserve full backwards compatibility

## Possible implementation

At a high level, this would likely involve:

1) adding a persisted STT base URL setting
2) exposing it in the Models / Settings UI alongside the existing provider configuration
3) reading the configured URL when creating STT provider clients
4) falling back to the current SDK defaults when the setting is empty

This keeps the feature opt-in while allowing Freestyle to work with OpenAI-compatible transcription servers.

i will begin implementing this under this issue

GatikTulsiani07:
## Summary

This PR adds backend support for configuring a custom base URL for OpenAI-compatible Speech-to-Text providers.

Previously, the OpenAI transcription provider always used the default OpenAI endpoint, making it impossible to use self-hosted or OpenAI-compatible STT servers. This change introduces an optional `openai_stt_base_url` setting that is validated, normalized, and forwarded to the OpenAI SDK while preserving the existing behavior when the setting is not configured.

## What changed

- Added `openai_stt_base_url` validation and normalization.
- Added backend support for configuring a custom OpenAI-compatible STT base URL.
- Preserved existing behavior when the setting is unset or empty.
- Added comprehensive tests covering:
  - default behavior
  - custom base URL
  - empty setting
  - URL normalization
- Added Models page UI for configuring the OpenAI STT custom base URL.
- Moved the configuration into the Models page (per review feedback), scoped only to the OpenAI STT provider.
- Removed explanatory code comments based on review feedback.

## Backend flow

An Excalidraw diagram has been attached to this PR illustrating:
Screenshot 2026-07-08 at 11 46 31 PM

1) The previous request flow where `createOpenAI({ apiKey })` always used the default OpenAI endpoint.
2) The new flow where `OpenAITranscriptionProvider` reads the optional `openai_stt_base_url` setting, normalizes it, and conditionally passes it as `baseURL` to `createOpenAI()`.
3) The unchanged execution path when no custom URL is configured, ensuring full backward compatibility.

## Testing

1)  Added dedicated unit tests for the new functionality.
2)  Verified default behavior is unchanged.
3) Verified custom endpoints are normalized correctly.
4) Verified empty settings fall back to the default OpenAI endpoint.
50  Ran the full server test suite (214 tests passing).

## Notes

This PR now includes both the backend implementation and the Models page UI requested during review.

The UI is scoped only to the OpenAI STT provider, persists the optional setting through the existing settings API, and preserves the default OpenAI behavior when left empty.

##Screenshots
PHOTO-2026-07-09-23-55-08

PHOTO-2026-07-09-23-55-23

Closes #415


matteo8p (changes requested):
1. Please leave code comments out of the PR. 
2. I'm not sure if the settings page is the correct place to add a custom base URL for speech to text models. I think the most sensible place to configure a model configuration is in the models page. Please consider adding or moving the UI to the models page. If you believe it belongs in the settings page, I would love to hear why. Feel free to push back on my opinions.

GatikTulsiani07:
Thanks for the review!

That makes sense. I agree that this is more of a model-specific configuration than a global application setting.

I'll move the UI to the Models page, keep the backend behavior unchanged, remove the code comments from the PR, and push an update shortly.

GatikTulsiani07:
hey @matteo8p i have bundled all the changes in this PR please review the changes sorry for taking so much time there were some merge issues as well so it took time to solve . Thanks Waiting for your review

MathurAditya724 (changes requested):
Reviewed the full diff. Backend logic is solid, tests are thorough (214 passing), and the UI mirrors `LocalLlmConnect` well for design-system consistency. Two things should be fixed before merge, plus a couple of minor notes.

## Must fix

**1. `zod` import diverges from the rest of the package** — `packages/validations/src/openai-stt.ts:1` imports `from "zod"` (v4), but every other file in this package imports `from "zod/v3"` (settings, models, local-llm, post-process, etc. — 13 files). `zod` is pinned at `^4.4.3`, so `zod/v3` is the deliberate compat entrypoint here. Mixing v3 and v4 schema instances can produce subtle type incompatibilities when composed and is an obvious inconsistency. Change to `import { z } from "zod/v3";`.

**2. Hand-rolled `URL` global declaration** — `openai-stt.ts:5-12` introduces a custom `ParsedUrl` interface plus `declare const URL: { new (input: string): ParsedUrl }`. Nothing else in the package does this; `settings.ts:56` (the `proxyUrlSettingSchema`, which is the direct analog) just calls `new URL(value.trim())` against the ambient global. Drop the `ParsedUrl`/`declare const URL` block and use `URL` directly to match the existing pattern. It builds fine via pkgroll either way, but the bespoke workaround adds noise and a maintenance trap.

## Minor / non-blocking

- **DRY with `proxyUrlSettingSchema`** — `openaiSttBaseUrlSchema` and `proxyUrlSettingSchema` are nearly identical (max 2048, http(s) refine, "or empty to disable" messaging). Not worth a shared helper yet, but worth noting the parallel structure if a third URL setting shows up.

- **`normalizeOpenAISttBaseUrl` throws inside `transcribe`** — `providers/openai.ts:15` calls the normalizer on the stored value. The server validates on write, and `transcribe` is wrapped in try/catch in `routes/transcribe.ts:240`, so a corrupt stored value degrades to a 500 rather than crashing — acceptable, just flagging the throw-in-hot-path.

- **Config visibility is tied to a visible OpenAI cloud row** — if the user filters/searches such that no OpenAI voice row is shown, the base-URL config disappears. Matches the "scoped to provider" intent, but means the setting can't be edited while filtered. Fine as-is.

GatikTulsiani07:
> Reviewed the full diff. Backend logic is solid, tests are thorough (214 passing), and the UI mirrors `LocalLlmConnect` well for design-system consistency. Two things should be fixed before merge, plus a couple of minor notes.
> 
> ## Must fix
> **1. `zod` import diverges from the rest of the package** — `packages/validations/src/openai-stt.ts:1` imports `from "zod"` (v4), but every other file in this package imports `from "zod/v3"` (settings, models, local-llm, post-process, etc. — 13 files). `zod` is pinned at `^4.4.3`, so `zod/v3` is the deliberate compat entrypoint here. Mixing v3 and v4 schema instances can produce subtle type incompatibilities when composed and is an obvious inconsistency. Change to `import { z } from "zod/v3";`.
> 
> **2. Hand-rolled `URL` global declaration** — `openai-stt.ts:5-12` introduces a custom `ParsedUrl` interface plus `declare const URL: { new (input: string): ParsedUrl }`. Nothing else in the package does this; `settings.ts:56` (the `proxyUrlSettingSchema`, which is the direct analog) just calls `new URL(value.trim())` against the ambient global. Drop the `ParsedUrl`/`declare const URL` block and use `URL` directly to match the existing pattern. It builds fine via pkgroll either way, but the bespoke workaround adds noise and a maintenance trap.
> 
> ## Minor / non-blocking
> * **DRY with `proxyUrlSettingSchema`** — `openaiSttBaseUrlSchema` and `proxyUrlSettingSchema` are nearly identical (max 2048, http(s) refine, "or empty to disable" messaging). Not worth a shared helper yet, but worth noting the parallel structure if a third URL setting shows up.
> * **`normalizeOpenAISttBaseUrl` throws inside `transcribe`** — `providers/openai.ts:15` calls the normalizer on the stored value. The server validates on write, and `transcribe` is wrapped in try/catch in `routes/transcribe.ts:240`, so a corrupt stored value degrades to a 500 rather than crashing — acceptable, just flagging the throw-in-hot-path.
> * **Config visibility is tied to a visible OpenAI cloud row** — if the user filters/searches such that no OpenAI voice row is shown, the base-URL config disappears. Matches the "scoped to provider" intent, but means the setting can't be edited while filtered. Fine as-is.

I’ve addressed both required items:

* Switched the validation import to zod/v3 to match the rest of the package.
* Removed the custom ParsedUrl / declare const URL workaround and aligned the implementation with the existing settings.ts pattern by using the global URL.

I also reviewed the three non-blocking suggestions:

* I kept openaiSttBaseUrlSchema separate because it intentionally differs from proxyUrlSettingSchema (protocol restrictions, query/hash validation, and /v1 normalization), so extracting a shared helper didn’t reduce complexity.
* The normalization throw path is already handled by the existing request error path, so invalid persisted values fail gracefully rather than escaping as unhandled errors.
* The Custom Base URL visibility is intentionally scoped to the visible OpenAI provider row, matching the current provider-specific UI design.

Please let me know if you’d still prefer a different direction for any of those points. 

Thanks!

MathurAditya724 (approved):
Re-reviewed after `e6ef475`. Both must-fix items are addressed:

- `openai-stt.ts` now imports `from "zod/v3"`, matching the rest of the package.
- Dropped the bespoke `ParsedUrl` interface and `declare const URL` shim; `parseHttpUrl` now returns `URL` and uses the ambient global directly, same as `proxyUrlSettingSchema` in `settings.ts`. Bonus: trimming moved into `parseHttpUrl`, so it's centralized.

Verified locally: validations builds via pkgroll, Biome clean, all 214 server tests pass. The remaining `tsc` `URL` type note is pre-existing and identical to `settings.ts` (package uses `lib: ["ESNext"]` and builds with pkgroll, not tsc), so it's consistent and out of scope.

LGTM.

GatikTulsiani07:
hey @matteo8p what do you think about this PR now ? Are there any further changes required for me to do before the merge

matteo8p (approved):
Ship it

MathurAditya724:
Just holding this PR for a bit, working on some cleanup. Will merge it after that

MathurAditya724:
Can you resolve the merge issues and can't we just have the flow for this like the Post processing models? 
Screenshot 2026-07-16 at 3 57 25 PM

MathurAditya724:
@GatikTulsiani07 let me know once you have completed the changes, will merge it.

Just wanted to know have you updated the UI

GatikTulsiani07:
hey @MathurAditya724 i have solved the merge conflicts and almost done with the UI changes as well , will update u in a bit after some checks

GatikTulsiani07:
hey @MathurAditya724 Bro i am done with the changes please have a look at it .

MathurAditya724:
Opened #479 as an alternative take on this, modeled on the existing on-device cleanup (local LLM) flow — adds a Test-connection button, simplifies validation, and gives the custom endpoint its own API key. Cross-linking for context.

MathurAditya724:
Thank you for this 🙏🏼 , created a new PR with minor tweaks inspired by this

Thank you for working on this

GatikTulsiani07:
Thanks @MathurAditya724  for the acknowledgment. I appreciate it. I do want to say that this was disappointing. After implementing all the requested changes and being told the PR would be merged once they were done, I didn't expect it to be closed in favor of another PR covering the same feature. If you don't mind, could you also share what specific technical or architectural reasons led to opening a new PR instead of continuing with this one? I'd genuinely like to understand what differences made that approach preferable so I can align my future contributions with the project's expectations. I also hope future contributors are informed earlier if the implementation direction changes, as it saves everyone a lot of time.

Thanks again, and I appreciate the feedback.

MathurAditya724:
Hey @GatikTulsiani07, fair question.

Your PR worked and the code was solid. Here's what we wanted to change and why we went with a new PR instead of more review rounds:

- Mirror the existing local-LLM connect flow exactly — same UI pattern, Test button that probes /v1/models, inline connection status
- Dedicated openai_stt_api_key instead of reusing the OpenAI key, since self-hosted gateways need their own credentials
- Simpler validation — drop the strict query/hash rejection and the throwing /v1 normalizer, store the URL verbatim
- Refactor the duplicated code — since the new STT flow now mirrors local-LLM, we extracted a shared component and hook so both use one implementation instead of two copy-pasted versions
- Unconditional top-of-list placement instead of the mid-list Fragment injection

That's a lot of structural changes across the hook, UI, validation, and a new test endpoint. We'd already asked you for multiple rounds of changes, and adding all of this on top felt unreasonable — it was easier to take the good parts from your PR and build on them in a new one than to try to explain all of this through more review comments.

Your backend flow and validation approach directly carried over into the final implementation. Appreciate the effort.

GatikTulsiani07:
Thanks  @MathurAditya724  for taking the time to explain—it makes a lot more sense now.

I appreciate the acknowledgment and the detailed breakdown of the architectural changes. Looking back, I can see why mirroring the Local LLM flow exactly and extracting the shared implementation was the cleaner long-term direction.

While it was definitely a little disappointing after going through the review rounds and implementing the requested changes, I appreciate the clarification and I'm glad the backend flow and validation work could still contribute to the final implementation.

Thanks again for the feedback. I learned a lot from working through this PR, and I'm looking forward to contributing more features and fixes in the future.

PR #449, opened 2026-07-12
fix: preserve right-side modifier-only hotkeys (RightCommand etc.)

### LINKED ISSUE #448: macOS: recording right ⌘ as hotkey saves generic Command — left ⌘ also triggers, and the app's own synthetic paste restarts recording
### Summary

Recording a **right-side modifier** (e.g. right ⌘) as a modifier-only push-to-talk hotkey silently saves the *generic* modifier category (`Command`) instead of the side-specific one (`RightCommand`). This causes two user-visible bugs on macOS:

1. **Both modifier keys trigger dictation.** The user picked right ⌘ only, but left ⌘ starts recording too, including during everyday shortcuts.
2. **Recording ghost-restarts right after hold-release.** On release, the transcript pastes correctly — but the pill closes and reopens a split second later and keeps recording. Freestyle is triggering itself: `macos-fast-paste` posts a synthetic Cmd+V (`macos-fast-paste.swift:60-64`), the key listener's CGEvent tap sees that event and emits `FLAGS:command` (`macos-key-listener.swift:402`), and a `Command` modifier-only hotkey matches it (`key-listener.ts` → `checkMacHotkeyMatch`). Nothing releases it afterwards, so recording continues until the next keystroke.

### Reproduction (v0.5.0, macOS arm64)

1. Settings → change the push-to-talk hotkey, press **right ⌘**. The saved value is `Command` (visible in `settings` table / the `macos-key-listener Command` process arg).
2. Press **left ⌘** → recording starts (bug 1).
3. Hold **right ⌘**, dictate, release → text pastes, then the pill immediately reopens and keeps recording (bug 2).

### Root cause

`hotkey-recorder.ts` deliberately collapses right modifiers to their generic category:

```ts
const RIGHT_MODIFIER_KEYS: Record<string, string> = {
  RightOption: "Alt",
  ...
  RightCommand: "Command",
  ...
};
```

so the settings UI can only ever save `Command`/`Alt`/… for a modifier-only capture — even though the runtime matcher **already fully supports side-specific hotkeys**: `checkMacHotkeyMatch` has a direct `macModState.has(...)` path for `RightCommand`/`RightOption`/etc., `normalizeAccelerator` and `isValidAccelerator` both accept them, and the native listener emits `RIGHT_MOD_DOWN/UP:RightCommand`.

I verified against the real matcher (driving `NativeKeyListener.handleLine` like `tests/key-listener.test.ts` does):

- hotkey `Command`: `FLAGS:command` alone fires `onKeyDown` → left ⌘ triggers, and the synthetic paste re-triggers (`down, up, down`).
- hotkey `RightCommand`: left ⌘ is ignored, right ⌘ hold/release works, and the synthetic Cmd+V does **not** re-trigger.

Manually writing `RightCommand` into the settings DB fixes both symptoms immediately, confirming the diagnosis end-to-end.

### Suggested fixes

1. **Recorder:** preserve side-specificity for modifier-only captures — when the recorded combo is just a right-side modifier, save `RightCommand`/`RightOption`/… instead of the generic category (chords like right-⌘+Space can keep collapsing to `Command+Space`, matching current matcher semantics).
2. **Self-trigger hardening (separate, smaller):** tag the synthetic paste events with a private `CGEventSource` user-data marker in `macos-fast-paste.swift` and skip them in `macos-key-listener.swift`, so even a deliberately-chosen generic `Command` hotkey can't be re-triggered by Freestyle's own paste.

Happy to submit a PR for both — wanted to file the diagnosis first in case you have preferences on the approach (especially whether modifier-only generic captures should remain possible at all).

symphonix84:
Fixes #448.

## What

Recording a right-side modifier (e.g. right ⌘) as a modifier-only push-to-talk hotkey now saves the side-specific accelerator (`RightCommand`) instead of silently collapsing to the generic category (`Command`).

With generic `Command` stored, the matcher fires on **either** ⌘ key, and — worse — on the app's own synthetic Cmd+V paste (`macos-fast-paste` posts V with `maskCommand`; the key listener's CGEvent tap emits `FLAGS:command` for it), so recording ghost-restarts right after every hold-release. Full diagnosis in #448.

The runtime already supports side-specific accelerators end-to-end (`NativeKeyListener` matching, `isValidAccelerator`, `normalizeAccelerator`, `isValidHotkeyCombo`); only the recording path lost the information — in two places (`hotkey-recorder.ts` `RIGHT_MODIFIER_KEYS` collapse, and the renderer hook's generic-only `orderModifiers`).

## How

Chord behavior is deliberately unchanged — right-⌘+Space still records as `Command+Space`, matching current matcher semantics. Only a **solo** right-modifier capture becomes side-specific:

- **`hotkey-recorder.ts` (main):** forward the side-specific token (`RightCommand`) to the renderer instead of pre-collapsing it.
- **`use-hotkey-recorder.ts` (renderer):** modifiers stay generic for display/merging, but a small latch (`nextRightModifierLatch`, pure + exported for tests) tracks "this draft is exactly one modifier, produced solely by a right-side key." On completion, a latched solo-modifier combo saves the side-specific accelerator.
  - The latch survives the trailing `FLAGS:command` the Swift binary emits from the same `flagsChanged` callback as `RIGHT_MOD_DOWN:RightCommand`, and the duplicate DOM delivery when the settings window is focused.
  - It clears on chords, on a captured non-modifier key, and on a definitively left-side DOM modifier press (`e.code === "MetaLeft"` etc.) — so pressing left ⌘ still records generic `Command`.
  - Display labels added so saved side-specific hotkeys render as `Right ⌘` / `Right ⌥` / … (mac) and `Right Alt` / `Right Super` / … (elsewhere) instead of raw `RightCommand`.

## Tests

`apps/electron/tests/key-listener.test.ts`: 20 passed (6 existing + 14 new), covering:

- `NativeKeyListener` with `RightCommand`: `FLAGS:command` alone (left ⌘ **or** the synthetic paste) never fires; `RIGHT_MOD_DOWN/UP` drives hold-to-talk correctly.
- `NativeKeyListener` with generic `Command`: existing behavior unchanged.
- `HotkeyRecorder`: `RIGHT_MOD_DOWN:RightCommand` now reaches the renderer side-specific.
- `nextRightModifierLatch`: realistic native event sequence (incl. the trailing `FLAGS:command`), DOM left-clear, chord-clear, and the ambiguous left-then-right case (stays generic).
- Pure helpers: `comboToAccelerator` / `isValidHotkeyCombo` with side-specific values.

Happy to split or adjust if you'd prefer a different approach (e.g. first-class right-modifier tokens in chords too). A separate follow-up could harden the listener against the app's own synthetic paste events (tagging them with a `CGEventSource` user-data marker) so even deliberately-generic `Command` hotkeys can't self-trigger — say the word and I'll send it.

### DIFF
--- a/apps/electron/src/main/hotkey-recorder.ts
+++ b/apps/electron/src/main/hotkey-recorder.ts
@@ -38,15 +38,6 @@ const MAC_RIGHT_MOD_KEYS: Record<string, string> = {
   rightshift: "RightShift",
 };
 
-const RIGHT_MODIFIER_KEYS: Record<string, string> = {
-  RightOption: "Alt",
-  RightAlt: "Alt",
-  RightCommand: "Command",
-  RightControl: "Control",
-  RightShift: "Shift",
-  RightSuper: "Super",
-};
-
 const MAC_FLAG_MODIFIERS: Record<string, string> = {
   control: "Control",
   option: "Alt",
@@ -245,8 +236,7 @@ export class HotkeyRecorder {
     if (line.startsWith("RIGHT_MOD_DOWN:")) {
       const modName = line.slice("RIGHT_MOD_DOWN:".length);
       const key = MAC_RIGHT_MOD_KEYS[modName.toLowerCase()] ?? modName;
-      const modifier = RIGHT_MODIFIER_KEYS[key] ?? key;
-      this.sendModifiers([...this.pendingModifiers, modifier]);
+      this.sendModifiers([...this.pendingModifiers, key]);
       return;
     }
 
--- a/apps/electron/src/renderer/src/hooks/use-hotkey-recorder.ts
+++ b/apps/electron/src/renderer/src/hooks/use-hotkey-recorder.ts
@@ -40,6 +40,31 @@ const KEY_SYMBOLS: Record<string, string> = {
   Fn: "\uD83C\uDF10",
   MouseButton4: "Mouse 4",
   MouseButton5: "Mouse 5",
+  RightAlt: "Right Alt",
+  RightOption: "Right Alt",
+  RightControl: "Right Ctrl",
+  RightShift: "Right Shift",
+  RightSuper: "Right Super",
+  RightCommand: "Right Super",
+};
+
+/** Side-specific right-modifier symbols shown on macOS (falls back to KEY_SYMBOLS elsewhere). */
+const MAC_RIGHT_MOD_SYMBOLS: Record<string, string> = {
+  RightControl: "Right \u2303",
+  RightCommand: "Right \u2318",
+  RightAlt: "Right \u2325",
+  RightOption: "Right \u2325",
+  RightShift: "Right \u21E7",
+};
+
+/** Maps a side-specific modifier token to the generic token it collapses to for display/merging. */
+const RIGHT_MODIFIER_TO_GENERIC: Record<string, string> = {
+  RightAlt: "Alt",
+  RightOption: "Alt",
+  RightCommand: "Command",
+  RightControl: "Control",
+  RightShift: "Shift",
+  RightSuper: "Super",
 };
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


MathurAditya724 (commented):
Reviewed the full diff, traced the recorder → renderer → matcher round-trip, and ran the suites locally on this branch.

**Verification**
- `pnpm --filter @freestyle-voice/electron test:e2e` → the new `key-listener.test.ts` passes 20/20. (The one `app.test.ts` failure is a headless-env `libglib-2.0.so.0` launch error, unrelated to this change.)
- `pnpm typecheck` clean, `biome check` clean, `knip` reports nothing on the new symbols.
- Confirmed the save/reload round-trip: `RightCommand` serializes via `comboToAccelerator`, `acceleratorToCombo("RightCommand")` returns `{modifiers:[], key:"RightCommand"}` (it's not in `MODIFIER_ORDER`/`MODIFIER_ALIASES`, so it's treated as the key), `isValidHotkeyCombo` accepts it via `MODIFIER_KEYS`, and `keyDisplayLabel` renders `Right ⌘`. So a saved side-specific hotkey re-renders correctly in Settings.

The diagnosis in #448 is solid and the fix is appropriately scoped — chords still collapse to generic, only a solo right-modifier capture becomes side-specific. The `nextRightModifierLatch` latch handling (trailing `FLAGS:command` survival, DOM `MetaLeft` clear, chord clear, left-then-right ambiguity) is well covered by the pure-function tests.

A few minor, non-blocking notes:

1. **Dead display fallbacks.** `KEY_SYMBOLS` (the non-mac branch of `keyDisplayLabel`) gained `RightAlt`/`RightOption`/`RightControl`/`RightShift`/`RightSuper`/`RightCommand`. But off macOS, side-specific tokens are never produced — `domRightModifierToken` returns `null` when `!IS_MAC`, and `RIGHT_MOD_DOWN` is emitted only by the macOS binary. So those entries are unreachable in practice. Not wrong, just dead weight; worth a comment noting they're defensive if you keep them.

2. **Two near-identical maps.** `RIGHT_MODIFIER_TO_GENERIC` duplicates `CAPTURED_MODIFIER_KEYS` (minus the `Fn` entry). Consider deriving one from the other to avoid them drifting apart.

3. **Native-only both-⌘ edge.** If someone holds right ⌘ and then also presses left ⌘ during recording, the DOM `MetaLeft` keydown sets `explicitLeft` and correctly clears the latch — good. The purely-native path (no DOM, settings window unfocused) can't distinguish sides on the generic `FLAGS:command` (count stays 1), so the latch would survive. In practice the settings window is focused during recording so the DOM handler covers it; just flagging it as a known narrow limitation, consistent with the scope you already called out.

Nothing here blocks merge. Nice, thorough work — the tests in particular make the tricky latch logic reviewable.

matteo8p (approved):
Ship it. Apologies for the late response. Thank you so much for putting in this PR! 

Also please consider joining our [Discord channel](https://discord.gg/Fmgt5yZCDu), that's where other contributors and I hang out.

PR #451, opened 2026-07-12
perf(electron): add bundle analysis and lazy-load i18n locales

MathurAditya724:
## Summary

Adds opt-in bundle analysis to the Electron app, fixes a pre-existing renderer build break, and lands the first optimization it surfaced (lazy-loaded i18n locales).

### 1. Bundle analysis tooling (`chore`)
- `rollup-plugin-visualizer` wired into all three `electron-vite` targets (main, preload, renderer), gated behind `ANALYZE=1`.
- New script: `pnpm --filter @freestyle-voice/electron run analyze` → writes a treemap per target to `stats/{main,preload,renderer}.html` (gitignored). Sourcemaps enabled only under analyze.

### 2. Fix zod v4 resolution (`fix`)
The renderer build was **broken on `main`**: `@hookform/resolvers@5.x` imports `zod/v4/core` but declares no zod dependency, so pnpm's isolated layout resolved it to a stray transitive `zod@3.x` (which has no `./v4/core` export) → `Missing "./v4/core" specifier in "zod"`. Pinned via `packageExtensions` in `pnpm-workspace.yaml` + a direct `zod` dep on the app. This also unblocks the normal `pnpm build`.

### 3. Lazy-load i18n locales (`perf`)
Locales were loaded via an **eager** `import.meta.glob`, bundling all 7 languages + `template.json` (~154 KB) into the initial dashboard chunk even though only one language is active. Now:
- Lazy glob (excluding `template.json`, which stays on disk as the contributor template).
- Minimal inline i18next backend resolves each language from its own chunk (no new dependency).
- Dashboard awaits `initI18n()` before mounting, so the active language is ready on first paint (no untranslated flash).

**Impact:** eager dashboard entry **868 KB → 700 KB** (~168 KB off initial load). Each language is now a ~17–22 KB on-demand chunk. `template.json` no longer bundled.

## Verification
- `typecheck:node` + `typecheck:web` clean
- Full `pnpm --filter @freestyle-voice/electron build` → exit 0
- `biome check` clean on changed files
- Analyze build confirms: 7 per-language chunks emitted, `template.json` absent, 0 locale bytes in the eager entry
- Language switching still works via the existing selector (backend auto-loads target chunk)

## Notes
- Pill window is unaffected (doesn't use i18n).
- Bundle analysis surfaced further opportunities not included here (markdown stack in plugin-detail ~280 KB lazy; three.js in the pill ~2 MB, isolated) — left for follow-up discussion.


MathurAditya724 (commented):
Reviewed and verified locally (approval blocked since this is my own PR).

- `typecheck:node` + `typecheck:web` clean
- `ANALYZE=1 build` exit 0
- biome clean on changed files (`i18n.ts`, `dashboard.tsx`, `electron.vite.config.ts`)

**Confirmed the PR's claims:**
- Eager renderer entry is **700 KB** (down from 868 KB), matching the stated ~168 KB reduction
- 7 per-language chunks emitted (`en`/`de`/`es`/`fr`/`it`/`ja`/`pt`), each ~18–24 KB, loaded on demand
- `template.json` is absent from the bundle

**Correctness notes:**
- The inline i18next backend's `read()` returns `module.default` for the `translation` namespace, matching the previous `{ translation: ... }` resource shape (no custom `defaultNS`), so lookups are unchanged.
- `initI18n()` resolves after the active language + `en` fallback load, so first paint is translated (no untranslated flash).
- Mount-on-failure fallback (`.catch(reportError).finally(mount)`) avoids a blank window if a locale chunk fails to load.
- zod v4 `packageExtensions` pin + direct dep correctly resolves the `zod/v4/core` break; the renderer build now completes.

No findings. LGTM.

PR #453, opened 2026-07-13
perf(electron): defer AI SDK loading and non-critical boot work

MathurAditya724:
## Summary

Two cold-start optimizations for the Electron app, found while tracing the startup path. Both cut work off the boot critical path without changing behavior.

### 1. Lazy-load AI provider SDKs (`perf(server)`)
`apps/server/src/lib/providers.ts` imported all five `@ai-sdk/*` packages at module scope. Because the route layer and the Electron main process both import the server at boot, **every provider SDK was eagerly evaluated on every cold start**, regardless of which provider the user uses.

Each SDK now loads via dynamic `import()` inside its factory. `createChatModel` (and its sole caller `resolveChatModel` in `post-process.ts`, already inside async `postProcess`) become `async`. Groq stays eager — it's the default cleanup provider and is already pulled in via `groq-http`.

**Effect:** eager main bundle **4.18 MB → 3.6 MB (~580 KB)**. The anthropic (209 KB), google (207 KB) and mistral (25 KB) SDK bodies are now lazy chunks loaded via `require()` behind an arrow function, only when that provider is actually used for cleanup.

### 2. Defer non-critical boot work (`perf(electron)`)
In the `whenReady` handler (`apps/electron/src/main/index.ts`):
- Wrapped `reconcileUnsupportedMlxVoiceDefault()` + `autoStartWhisperServer()` in `setImmediate(...)`. `reconcile` can synchronously `execFileSync` a Python/MLX probe (5 s + 30 s timeouts) on Apple Silicon without a managed runtime, blocking first paint. Both are idempotent and also run lazily via `getDefaultModels()` on first use, so deferring a tick is safe.
- Added `AbortSignal.timeout(1500)` to the startup `/api/health` probe so a half-open socket on port 4649 can't hang window/tray creation. The existing `catch` handles the abort.

## Verification
- `typecheck:node` + `typecheck:web` clean (server code is covered by the web tsconfig)
- **202/202 server tests pass**
- Full electron `build` exit 0
- biome clean
- Confirmed the deferred SDKs compile to lazy `require()` and are out of the eager bundle

## Notes
- Independent of #451 (renderer bundle analysis + lazy locales) — no file overlap.
- On local disk this is primarily reduced eager parse/eval work + not blocking the main thread; the biggest remaining startup heavyweight is the ~2 MB three.js pill chunk, left for a separate change.

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -1955,9 +1955,16 @@ app.whenReady().then(async () => {
     process.env.FREESTYLE_MLX_ASR_RELEASE_TAG ||= app.getVersion();
   }
 
-  // Run non-critical server startup tasks now that the DB path is set
-  reconcileUnsupportedMlxVoiceDefault();
-  autoStartWhisperServer();
+  // Run non-critical server startup tasks now that the DB path is set. These
+  // are deferred off the boot critical path: reconcileUnsupportedMlxVoiceDefault
+  // can synchronously probe Python/MLX (execFileSync) on Apple Silicon without a
+  // managed runtime, which would otherwise block window creation. Both are
+  // idempotent and also run lazily via getDefaultModels() on first use, so
+  // deferring them by a tick is safe.
+  setImmediate(() => {
+    reconcileUnsupportedMlxVoiceDefault();
+    autoStartWhisperServer();
+  });
 
   // Start the Hono HTTP server with WebSocket support (or reuse an existing one)
   const startServer = (port: number): void => {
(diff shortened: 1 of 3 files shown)


MathurAditya724 (commented):
Reviewed and verified locally (approval blocked since this is my own PR).

- `typecheck:node` + `typecheck:web` clean
- **202/202 server tests pass**
- biome clean on changed files (`main/index.ts`, `post-process.ts`, `providers.ts`)

**Correctness notes:**
- Traced the async propagation: `createChatModel` is now `async` and its sole caller `resolveChatModel` (in `post-process.ts`) is `async` and correctly `await`ed inside `postProcess`. `getGroqChatModel` stays synchronous, and the benchmark/tone scripts that call it directly are unaffected.
- Lazy `import()` inside each provider factory keeps the anthropic/google/mistral SDK bodies out of the eager bundle; Groq stays eager as the default cleanup provider.
- `setImmediate(...)` deferral of `reconcileUnsupportedMlxVoiceDefault()` + `autoStartWhisperServer()` is safe — both are idempotent and also run lazily via `getDefaultModels()` on first use, so moving them off the boot critical path won't drop behavior.
- `AbortSignal.timeout(1500)` on the startup `/api/health` probe is handled by the existing `catch`, so a half-open socket can't hang window/tray creation.

No findings. LGTM.

PR #455, opened 2026-07-13
refactor(settings): split settings page into per-tab react-hook-form components

MathurAditya724:
## Summary

Breaks the 1515-line `pages/settings.tsx` monolith (~22 `useState` in one component) into a `pages/settings/` folder mirroring the existing `models/` convention. **Each tab is now its own react-hook-form**, while keeping the current **persist-on-change** behavior (no Save buttons — every change still writes immediately via `PUT /settings/:key` or `window.api.*`).

## Why

The page mixed three unrelated concerns in one giant component: persisted settings, IPC-backed toggles, and transient UI (device enumeration, updater events, permission polling). It was hard to navigate and extend. The `NetworkPanel` already proved the RHF + persist-on-change pattern works well; this generalizes it.

## What changed

**New `apps/electron/src/renderer/src/pages/settings/`**
- `index.tsx` — shell: title, update banner, sidebar, hash-synced section routing
- `use-settings-form.ts` — **core**: shared persist-on-change engine (dedup unchanged values → `trigger()` validate → `PUT`/IPC → patch `["settings-all"]` cache → optional "Saved" flash)
- `recording-panel.tsx`, `application-panel.tsx`, `display-panel.tsx`, `data-panel.tsx`, `network-panel.tsx` — one RHF form per tab
- `permissions-panel.tsx` — intentionally form-less (no editable fields; only shows OS status + triggers native dialogs)
- `components.tsx` — shared `Row`, `SettingsPanel`, `Segment`, `PermissionControl`, `SettingsSidebar`
- `constants.ts`, `use-app-updater.ts`, `use-permissions.ts` — extracted constants + transient-state hooks

**`packages/validations/src/settings.ts`**
- Centralizes previously-inline string unions as enums (`hotkeyModeSchema`, `outputModeSchema`, `themeSchema`, `audioPlaybackModeSchema`, `historyRetentionPresetSchema`) and adds per-tab form schemas (`recording/application/display/dataSettingsFormSchema`) with the same rules the server enforces.

## Behavior preservation (no functional change)

- Each panel reads via `useQuery(settingsQueryOptions())` — React Query dedupes into the single shared `["settings-all"]` cache, so splitting adds **no** extra fetches.
- Dual-key `audio_playback_mode` + `audio_ducking_enabled` write, legacy audio fallback chain, `sendOutputModeChanged` / `setHotkeyMode` / `sendAudioPlaybackModeChanged` side-effects, custom-retention validation/clamping, and Network trim + saved-flash are all preserved.
- IPC-backed settings (auto-update, launch-at-startup, show-on-launch, pill position, theme) seed from `window.api` reads; transient state (device list, updater events, section nav, hotkey recorder) stays outside RHF.

## Verification

- `typecheck:node` + `typecheck:web` — clean
- `apps/server` + `packages/validations` — clean
- `biome check` — clean
- `knip` — clean
- `electron-vite build` — succeeds

## Note

Permissions stays form-less by design — it has zero editable fields, so wrapping it in `useForm` would add ceremony with no validation/submit benefit.


MathurAditya724 (commented):
Reviewed and verified locally (posting as a comment — GitHub blocks self-approval on own PRs).

- `typecheck:node` + `typecheck:web` — clean
- `packages/validations` + `packages/stt` build — clean
- `electron-vite build` — exit 0
- `biome check` on all 13 changed files — clean
- `knip` — clean (only pre-existing config hints, nothing from this PR)

**Behavior preservation (the core risk for a "zero functional change" refactor):**

I cross-checked every settings key and side-effect in the deleted 1515-line `settings.tsx` against the new `settings/` folder. **Every key is accounted for — no dropped settings, no missing side-effects.** Specifically confirmed:

- Dual-key audio write (`audio_playback_mode` + `audio_ducking_enabled`) preserved, with `sendAudioPlaybackModeChanged` firing in the `after` hook (recording-panel L423-435).
- Legacy audio fallback seeding chain (`audio_playback_mode` → `pause_playback_while_recording` → `audio_ducking_enabled`) reproduced exactly.
- `sendOutputModeChanged` / `setHotkeyMode` side-effects preserved via per-field `after` hooks.
- History-retention clamping (digit-strip, 4-char slice, `HISTORY_RETENTION_DAYS_MAX` clamp) + "save only if valid" now enforced by the `dataSettingsFormSchema` refine through `trigger()`.
- Network trim + saved-flash + dedup preserved in the shared `useSettingsForm` engine.
- All IPC-backed seeds (auto-update, launch-at-startup, show-on-launch, pill position, theme), hash-synced section routing, and the permissions polling (1s interval / 30s timeout) moved faithfully.

**Notes (not blocking):**
- The new per-field dedup means clicking an already-active segment no longer issues a redundant `PUT` (old code always wrote). Not user-observable since `SegmentedControl` doesn't fire on same-value clicks.
- Seeding now normalizes out-of-enum stored values (e.g. a malformed `output_mode`) to the default rather than passing them through raw — effectively a small hardening, consistent with intent.
- The secondary `audio_ducking_enabled` write isn't patched into the `["settings-all"]` cache, but neither was it in the old code — it's a legacy-compat flag only.

The `packages/validations/src/settings.ts` additions are purely additive (new enums + per-tab form schemas reusing existing `parseRetentionDays`/`HISTORY_RETENTION_DAYS_MAX`); no existing export changed, so the server side is unaffected.

No findings. Clean, faithful refactor. LGTM.

MathurAditya724:
Closing without merging.

After review, this refactor doesn't carry its weight:

- **No user-facing value** — it's a pure no-functional-change restructure (verified: every settings key and side-effect preserved).
- **Low maintainability payoff** — the settings page is a stable area, so the "easier to extend" benefit of splitting the monolith mostly doesn't materialize.
- **Nothing standalone to salvage** — the new per-tab zod schemas in `packages/validations/src/settings.ts` are form-type contracts consumed only by this PR's own panels. The server validates a disjoint set of keys (`cleanup_*`, `plugins`, proxy, keep-alive) and doesn't touch `hotkey_mode`/`output_mode`/`theme`/`audio_playback_mode`/`history_retention_days`, so there's no client/server validation-centralization to cherry-pick — the schemas would be dead code without the refactor.

If settings becomes an active area again, this is worth revisiting. A separate, higher-value follow-up would be to add real server-side validation for those settings keys using shared enums — that would create the shared client/server validation this PR was assumed (incorrectly) to provide.

PR #459, opened 2026-07-13
refactor(server): unify cloud model adapters behind provider registries

MathurAditya724:
## Summary

The per-provider "cloud model" adapters for STT and LLM cleanup were quite similar / duplicated. This consolidates them behind two consistent, data-driven registries so adding a provider is a single descriptor.

**Scope (per discussion):** internal refactor only, both STT + LLM, **dispatch layer only** — no plugin/third-party registration seam, and model catalogs, key handling, and UI are untouched. **No behavior change.**

## STT / transcription
- New `makeAiSdkTranscriptionProvider()` in `streaming/utils.ts` for any Vercel AI SDK provider (`createX({ apiKey }).transcription(id)`), with optional `modelTransform` + `biasHandler` escape hatches.
- `streaming/registry.ts` is now a descriptor array. Deleted the byte-identical `providers/groq.ts` / `providers/openai.ts` and the `elevenlabs.ts` boilerplate — ElevenLabs' `_realtime` rewrite and keyterm-bias handling are now inline options.
- `deepgram`, `freestyle-cloud`, `whisper-local`, `mlx-local` keep their dedicated classes (bespoke transports).

## LLM cleanup / post-processing
- New `lib/llm/registry.ts`: an `LlmProvider` interface (`createModel`, optional `providerOptions`/`prewarm`, `local` flag) + descriptor array + `getLlmProvider()` with exact-then-prefix lookup. Groq's cached keep-alive model, reasoning provider-options, and prewarm now live in its descriptor. SDK imports stay lazy.
- `providers.ts` `createChatModel()` dispatches through the registry; removed `PROVIDER_FACTORIES` / `findFactory`.
- `post-process.ts` drops the groq special-cases in `resolveChatModel`, provider-options, and prewarm. `groqCleanupProviderOptions` is re-exported for the existing tests/benchmark scripts.
- Demoted `groqFetch` to a module-local const (no longer imported externally).

## Net effect
Adding a cloud wrapper is now one entry: a `makeAiSdkTranscriptionProvider(...)` line in the STT registry and/or an `LlmProvider` object in the LLM registry. Provider special-casing no longer leaks into `post-process.ts`.

## Verification
- `tsc --noEmit` clean
- `biome check` clean
- `vitest run` — 202/202 tests pass (incl. `post-process-options` and `groq-http`)

## Out of scope / possible follow-ups
- Plugin seam for third-party provider registration.
- Contributing models to the picker UI + key validation for new providers.
- Optional cleanup: unify the 3 prefix-strippers and the two provider-category helpers.


MathurAditya724 (commented):
Reviewed locally against `main`. Verified the refactor is behavior-preserving:

- **STT dispatch** — `makeAiSdkTranscriptionProvider` reproduces the deleted openai/groq/elevenlabs classes exactly. For openai/groq (no options) it's a straight passthrough to `transcribeWithAiSdk`, which still does the `stripProviderPrefix`. ElevenLabs' `modelTransform` (`_realtime` strip) + `biasHandler` (keyterm escape hatch) match the old inline logic, including operand order (transform then bias). `if (biased) return biased` is safe since `transcribeElevenLabsWithBias` always resolves a truthy result.
- **LLM dispatch** — `getLlmProvider`'s exact-then-prefix lookup preserves the old `findFactory` order (openai, groq, anthropic, google, mistral, local-llm); no provider id is a prefix of another, so no collisions. Groq's cleanup path routes through the cached keep-alive `getGroqChatModel` exactly as the old `resolveChatModel` did — `createChatModel` is the sole caller (`post-process.ts:332`), so no other site changes behavior.
- **Re-export chain** — `groqCleanupProviderOptions` moves to `llm/registry.ts` and is re-exported from `post-process.ts`; the two tests and two benchmark scripts still import it from `post-process.js`. Confirmed no dangling references to any removed symbol (`PROVIDER_FACTORIES`, `findFactory`, `resolveChatModel`, the three deleted provider classes, or the demoted `groqFetch`).

Checks reproduced locally: `tsc --noEmit` clean, `biome check` clean, `knip` clean (only pre-existing config hints), `vitest run` 202/202 pass.

One nit, already noted in the PR as an out-of-scope follow-up: `stripGroqPrefix` in `llm/registry.ts` duplicates `normalizeGroqModelId` in `groq-http.ts`. Fine to leave for the prefix-stripper unification pass. No blocking findings.

PR #462, opened 2026-07-13
feat(plugins): overhaul plugin pipeline with cancellation, capture hooks, and LLM capability

MathurAditya724:
## Summary

Redesigns the plugin hook model to close control/capability gaps (cancellation & suppression, a capture/audio-stage hook, LLM access for plugins, extended cleanup control) **and** consolidates the plugin system to be server-owned end to end — including serving plugin UI from the server, so the Electron app keeps only OS-level responsibilities.

This is a **breaking SDK change** (major bump). Both shipped plugins (`profanity-filter`, `audio-transcription`) work unchanged against the new SDK.

## What's in this PR

**SDK (`packages/sdk`)** — breaking `Handler` signature (additive at call sites; 2-arg handlers keep working):
- `HookApi`/`PipelineControl`: first-class `consume()`/`abort()`/`stopPropagation()` instead of the old empty-string-to-suppress convention.
- New `beforeTranscribe` hook: audio/provider/model/language/bias override before STT.
- New `PluginLlm` capability (`api.llm`) — plugins call the host's configured LLM without ever seeing API keys.
- Extended `beforeCleanup`: `prompt` replacement and `skip`.
- `PluginContributes.settings` **types + parser** (`parsePluginSettingsFields`) — SDK-only for now; host-side rendering + a validations schema are a follow-up.
- `registry.has(hookName)` for host-side hook-presence detection.
- `FreestyleBridge.api()` now resolves the native `Response` (plugin UI is same-origin with the server).

**Server (`apps/server`)**:
- `transcribe.ts` threads a per-dictation `HookApi` through `beforeTranscribe → afterTranscribe → beforeCleanup → afterCleanup`. Terminal control state (`consume`/`abort`) blanks the output on every return path so delivery is actually suppressed.
- Adaptive cloud mode: falls back from combined STT+cleanup to raw STT + local cleanup when a plugin needs `afterTranscribe`/`beforeCleanup`.
- `post-process.ts`: `beforeCleanup`'s `prompt`/`skip` wired in; a consumed/aborted dictation skips the LLM call.
- New `POST /api/output/deliver`: runs `beforeOutput` server-side on the final (post multi-segment-merge) text; honors `consume()`/`abort()`.
- New `POST /api/events`: relays app-originated events into the server's single `event` sink.
- New `GET/PUT/DELETE /api/plugins/:name/storage/:key` for plugin UI pages.
- **New `GET /api/plugins`** (hub list) and **`GET /api/plugins/:slug/ui/*`** (serves plugin UI assets, path-traversal guarded, with a server-set path-scoped CSP). Discovery + asset resolution live in `lib/plugins/ui-assets.ts`.

**Electron (`apps/electron`)**:
- Deleted the app-side plugin hook registry, and now the entire UI-hosting layer: the `freestyle-plugin://` custom scheme (`ui.ts`), app-side discovery (`manifest.ts`), the HTTP wrappers (`loader.ts`), and the `plugin-bridge:fetch` proxy protocol (`bridge-protocol.ts`).
- `view-manager.ts` loads pages same-origin from the loopback server, each in its own `session` partition (so one plugin can't read another's storage/cookies despite the shared origin).
- `ui-host.ts` trimmed to the view overlay + host-action IPC.
- `preload/plugin-bridge.ts`: `window.freestyle.api()` is a plain same-origin `fetch`; `invoke()` (host actions) unchanged.
- The renderer plugins hub/detail/page talk to the server directly over the typed `hc` client (`renderer/src/lib/plugins-api.ts`) instead of `window.api.*` IPC.
- `deliverOutput()` is purely mechanical; recording/output events relay to `POST /api/events`.

## Follow-up

Remote-server UI-asset auth (a `FREESTYLE_AUTH_TOKEN` for non-loopback servers). The loopback/local path that ships in the desktop app is fully covered here.

## Verification

- SDK: `pnpm build` + `pnpm test` — 19 tests.
- Server: `tsc --noEmit` clean; `pnpm test` — 216 tests (new `output-and-events.test.ts`, `output-control.test.ts`, `plugin-ui.test.ts`).
- Electron: `typecheck:node` + `typecheck:web` clean; `electron-vite build` succeeds (main/preload/renderer).
- `biome check` and `knip` clean.
- Shipped plugins (`profanity-filter`, `audio-transcription`) typecheck unchanged against the new SDK.

> Interactive Electron verification (opening both plugin pages, host actions, per-plugin session isolation) is still pending before this leaves draft.


MathurAditya724:
## Deep self-review

Read the spec + all 27 files and ran full verification. The PR's original claims hold: SDK build + 19 tests, server `tsc --noEmit` clean + 209 tests, electron `typecheck:node`/`typecheck:web` clean, biome + knip clean. Architecture and SDK-level control primitives are solid and well-tested.

Found one significant correctness gap plus doc/scope issues. **All fixed in `b1a2b06`.**

### 🔴 Bugs (fixed)

**1. `control.consume()`/`abort()` only suppressed delivery when the plugin *also* emptied the text — the transcribe `disposition` was dead.**
The renderer read `disposition` only from `/api/output/deliver`, never from `/api/transcribe` (`TranscribeResult` had no such field). So:
- `afterTranscribe` consume → worked only by accident (the raw-stage guard at `transcribe.ts:398` early-returns empty text).
- `beforeCleanup`/`afterCleanup` consume/abort → **broken**: those run inside `postProcess`, past that guard; the final return sent non-empty `cleaned` with `disposition: "suppressed"` and the renderer pasted it. Spec §5.4's own voice-command example (`consume()` without emptying text) would get pasted.
- cloud-combined path never checked `control.state` at all.

Fix: server now blanks output on terminal control state across **every** return path (local + cloud-combined); renderer treats a non-`"deliver"` disposition as suppressed via `isDeliverable`.

**2. `beforeOutput` ignored `consume()`/`abort()`.** `output.ts` derived `disposition` from mode/text first and only consulted `control.state` to *label* an already-suppressed result. A `beforeOutput` handler that consumed/aborted but left `mode:"paste"` + text still pasted. Fix: the deliver route now short-circuits on `control.state !== "running"` directly.

### 🟡 Docs / scope (fixed)

**3.** Spec §11 claimed `post-process-route.ts` "runs `afterTranscribe` too (symmetry)". It doesn't — and shouldn't: each segment already ran `afterTranscribe` in its own `/api/transcribe` call, so re-running on merged multi-segment text would double-fire. Spec corrected to document the deliberate choice.

**4.** Declarative settings was listed as "Shipped" but is inert: `PluginSettingField` + `parsePluginSettingsFields` exist and are tested in the SDK, but nothing consumes them and the `packages/validations` Zod schema (§8.2/§11) was never added. Moved host-side rendering + the Zod schema to **Deferred**; only the SDK types + parser ship here. PR body + spec §0/§8.2/§14 updated.

**5.** `.lore.md` carried unrelated lore-distillation churn (STT registry, groq, workflow-entry deletions from other sessions). Reverted to `main` — lore re-derives from committed work.

### 🟢 Test gap (fixed)
No end-to-end test covered "consume/abort actually suppresses delivery" — exactly where the bugs lived. Added `tests/output-control.test.ts` (3 tests) covering consume/abort suppression through `/api/output/deliver` with text/mode left intact.

### Verification after fixes
SDK 19 ✓ · server `tsc` clean + **212** tests (+3) ✓ · electron typecheck ✓ · biome ✓ · knip ✓

MathurAditya724:
## Consolidation landed (server-owned plugin UI)

Per the discussion, pulled the deferred UI-hosting slice into this PR and removed the spec file.

**Server now owns everything:** discovery, install/uninstall, catalog, updates, hooks, storage, **and UI serving**. Added `GET /api/plugins` (hub list) + `GET /api/plugins/:slug/ui/*` (assets, path-traversal guarded, server-set path-scoped CSP) in `lib/plugins/ui-assets.ts`.

**Electron is now a thin executor:** deleted the `freestyle-plugin://` custom scheme (`ui.ts`), app-side discovery (`manifest.ts`), the HTTP wrappers (`loader.ts`), and the `plugin-bridge:fetch` proxy (`bridge-protocol.ts`). `view-manager.ts` loads pages same-origin from the loopback server, each in its own `session` partition to preserve cross-plugin isolation. `window.freestyle.api()` is now a plain same-origin `fetch`; host actions (`invoke`) stay on IPC.

**Renderer** talks to the server over the typed `hc` client (`lib/plugins-api.ts`) instead of `window.api.*` plugin IPC.

Net **−1094 lines**. Spec file removed (`specs/` convention aside, the design is captured in the PR description now).

**Deferred (only remaining piece):** remote-server UI-asset auth (`FREESTYLE_AUTH_TOKEN` for non-loopback servers). The local/loopback path that ships in the app is fully covered.

**Verified:** SDK 19 · server tsc clean + 216 tests (new `plugin-ui.test.ts`) · electron typecheck + `electron-vite build` · biome + knip clean.

⚠️ Still needs interactive Electron verification (both shipped plugin pages, `copy`/`navigate` host actions, per-plugin session isolation) before leaving draft — that can't be done headlessly.

MathurAditya724:
## Deep review pass — findings fixed (`2d14982`)

Reviewed the full branch (all 41 files) with two independent read-passes. Found and fixed the following; everything verified green (SDK 20 · server tsc + 221 tests · electron typecheck + `electron-vite build` · biome + knip).

### Bugs
1. **`afterCleanup` ran on a consumed/aborted pipeline.** `registry.run` only checked per-hook `propagationStopped`, not terminal `control.state`, so a `consume()` in `afterTranscribe` didn't stop the later `afterCleanup` stage (contradicting the `consume()` "skip all remaining stages" contract). Fixed at the root: `run()` now returns early when `state !== "running"`. Removes reliance on every caller re-checking state.
2. **App→server `pipelineError` relay was silently 400'd.** `deliverOutput` relays a `pipelineError` on paste failure, but `/api/events`'s enum only allowed the 4 recording/output types. Rewrote the schema as a `discriminatedUnion` matching `FreestyleEvent` (incl. `pipelineError` + `outputDelivered` with required fields) and dropped the unsafe `as FreestyleEvent` cast.
3. **No `outputDelivered` event on suppression.** When `/deliver` returns a non-`deliver` disposition the renderer skips paste/copy, so `deliverOutput` never emitted. `/api/output/deliver` now emits `outputDelivered{none}` (suppressed) / `pipelineError` (aborted) server-side — the paths Electron never reaches — while delivery stays Electron-emitted (no duplicate).
4. **`abort()` never emitted `pipelineError`** despite the SDK doc promising it. Added `emitAbortEvent` on every aborted terminal path in `transcribe.ts` + `post-process-route.ts` + `output.ts`.
5. **Multi-segment `consume/abort` was lost.** `/api/post-process` now returns `disposition` (and blanks text); the renderer honors it instead of falling back to the combined raw.

### Edge cases / perf
6. **`beforeOutput` resolved an LLM per delivery** even though the SDK contract says it never receives one → new `createOutputHookApi()` skips it.
7. **`plugins-api.ts` dropped request timeouts** the old loader had → restored 5s/60s bounds; `setPluginEnabled` now checks `.ok` before claiming success.
8. **`beforeTranscribe` `bias` override was ignored on the cloud STT path** → now honored there too.

### DRY / dead code
- `dispositionFromControl` + `Disposition` moved to `pipeline.ts`, shared by `transcribe.ts`/`output.ts`/`post-process-route.ts` (was duplicated inline in output.ts).
- Extracted a `suppressedResponse()` helper (the blank-output JSON was repeated 3×).
- `parseAppContext` was computed twice in one object literal → cached once.
- Removed dead `parseAppContext` re-export in electron `plugins/index.ts`; added `.wasm`/`.avif` MIME types.

### Reviewed & OK (no change)
Route ordering (no Hono collision), path-traversal guard (string logic correct; symlink-escape noted as trusted-author risk), per-plugin session partition, `PluginInfo`↔server shape match, `app.tsx` `isDeliverable` filtering, no dangling `freestyle-plugin://`/`plugin-bridge:fetch`/`FreestyleResponse` references.

### Known follow-ups (unchanged, intentional)
Remote-server UI-asset auth token; discovery re-runs per asset request (fine for small UIs opened rarely); interactive Electron verification still pending before leaving draft.

MathurAditya724:
## Merged latest `main` — conflicts resolved (`ff0679c`)

`main` had landed three overlapping plugin PRs (#463 sidebar nav, #464/#465 cloud-fetched catalog). Merged and resolved:

**Conflicts (4 files):**
- `routes/plugins.ts` — kept my new `GET /` + `GET /:slug/ui/*` (asset serving) alongside main's cloud-proxied `/catalog`; dropped the deleted `catalog.ts`/`PLUGIN_CATALOG` import.
- `main/plugins/view-manager.ts` — kept main's **multi-view cache** (`views` Map + `activeKey`, instant page-switch) and adapted it to my server-URL loading + per-plugin `session` partition + tokens-only config (main's version still used the deleted `freestyle-plugin://` scheme).
- `plugins/plugins.tsx` — kept main's `placeholderData: keepPreviousData`; kept my `plugins-api` client calls.
- `plugins/plugin-detail.tsx` — kept main's `Switch`-based layout; dropped the now-unused `DropdownMenu` import; kept my `plugins-api` calls.

**Cross-file breakage from the merge (fixed):**
- `shell.tsx` (new sidebar nav) called the removed `window.api.refreshPlugins()` → now uses `listPlugins()`.
- `getPluginCatalog()` → the cloud `/catalog` body is untyped, so asserted the `{ plugins }` shape.

**Verified post-merge:** SDK 20 · server tsc + 221 tests · electron typecheck (node+web) · `electron-vite build` · biome + knip clean. PR is now `MERGEABLE`.

PR #467, opened 2026-07-15
feat: restore streaming audio behind experimental flag

MathurAditya724:
## Summary

Restores the real-time WebSocket streaming pipeline removed in PR #410, gated behind a new **Streaming Audio** toggle in **Settings > Experimental**. When disabled (default), the existing batch WAV POST flow is completely unchanged.

## What streaming does

Instead of recording the full audio and sending it as a WAV after you stop speaking, streaming sends audio chunks to the transcription provider in real-time via WebSocket as you speak. By the time you stop, 95%+ of the transcript is already done.

**Latency improvement:**
| Duration | Batch (current) | Streaming (new) |
|----------|----------------|-----------------|
| 30s dictation | ~3-4s wait | ~2-3s wait |
| 2min dictation | ~6-8s wait | ~3s wait |

## Supported providers

| Provider | Transport | Notes |
|----------|-----------|-------|
| Freestyle Cloud | WSS → `/v2/stream` → Soniox DO | Cloud handles STT + cleanup |
| OpenAI (BYOK) | WSS → `api.openai.com/v1/realtime` | Live partials, `gpt-4o-transcribe` |
| Deepgram (BYOK) | WSS → `api.deepgram.com/v1/listen` | Live partials, `nova-3` |
| ElevenLabs (BYOK) | WSS → `api.elevenlabs.io` | Live partials, `scribe_v2_realtime` |
| Soniox (BYOK) | WSS → `stt-rt.soniox.com` | Live partials, `stt-rt-v4` |
| Groq / Whisper / MLX | N/A | No streaming support — falls through to batch |

### Settings & Flag
- New `streaming_audio` setting key
- New **Experimental** panel in Settings with streaming toggle
- i18n key added for section label

### Server (apps/server)
- Restore streaming types (`StreamSession`, `StreamCallbacks`, `StreamingSessionOptions`, `StreamCleanupPreferences`)
- Restore 6 streaming providers (OpenAI, Deepgram, ElevenLabs, Soniox, Freestyle Cloud, MLX)
- Restore `stream.ts` route (773 lines) — **gated**: returns 400 if `streaming_audio !== "true"`
- Add `freestyleCloudStreamWsUrl()` helper
- Restore `pcm.ts`, `segments.ts`, `session-policy.ts`
- Add `soniox-context` to `AsrVocabularyBias` union
- Add `streaming` / `streaming_handoff` to `PostProcessSource`
- Add `ws` + `@types/ws` dependencies
- Restore Soniox model in `BUILTIN_VOICE_MODELS`

### Renderer (apps/electron)
- Restore `Streamer` class (WebSocket + AudioWorklet PCM capture)
- Restore `pcm-processor.ts` (AudioWorkletProcessor)
- Restore `encodeWavFromInt16` in `wav.ts`
- Add streaming integration in `app.tsx` — **all gated**: streamer only created/used when `streaming_audio` is `"true"`
- Add streaming badge + Soniox to models metadata
- Add Experimental settings panel

### Cloud repo
**Zero changes needed.** The `/v2/stream` endpoint and `SttSession` Durable Object are already live and operational.

## Verification
- `tsc --noEmit`: clean (server + electron node + electron web)
- `biome check`: clean
- `turbo build`: 9/9 tasks pass
- `vitest run`: 209/209 tests pass (24 files)

## Files changed
38 files changed, 3,281 insertions, 112 deletions


MathurAditya724 (commented):
## Review — restore streaming audio behind experimental flag

Verified locally: server tests pass (209/209, 24 files). Reviewed the full diff against `main` with two focused read passes (server + renderer).

**Gating is airtight and the default-off batch path is genuinely untouched.** The flag gate in `stream.ts` runs before `upgradeWebSocket`, so a client can't reach streaming logic with the flag off — the upgrade is rejected with 400 at the node-server layer. `routes/transcribe.ts` is not modified; all other batch-adjacent changes are strictly additive (enum values, a Soniox model entry, a `soniox-context` bias kind). When `_streamingAudioEnabled` is false the Streamer is never constructed and every `streamerRef.current?.cancel()` is a no-op.

The concerns below are all on the **flag-on** path.

### Blocking (for enabling the experimental path)

**1. Streamer is never destroyed — WebSocket + AudioContext leak for the app lifetime.**
`streamer.ts` reconnects 1s after any close while `sessionTransportSupported` (lines 265-272). But `streamer.destroy()` is never called anywhere — unmount cleanup only destroys the Recorder (`app.tsx:1183`). Once created, the WebSocket reconnect loop and the `AudioContext`/worklet persist for the whole app lifetime. Add `streamerRef.current?.destroy()` to the unmount path.

**2. Single-slot `streamResolverRef` mismatches transcripts under re-record.**
The re-record gate (`app.tsx:1134-1141`) allows starting a new recording while `transcribing`. In the streaming path, the second commit overwrites `streamResolverRef.current` (`app.tsx:895`), so session #1's `final` resolves into session #2's slot — transcripts get mismatched/dropped. The Streamer is also a single WebSocket + single PCM buffer and can't represent two concurrent sessions. Either serialize streaming commits or key the resolver per session.

**3. Destructive shared WAV buffer makes REST fallback unreliable under re-record.**
`getWavBlob()` clears the buffer (`streamer.ts:131-141`) and `startCapture()` wipes `pcmChunks` (lines 73-74). If session #2 starts capture before session #1's 15s timeout fires its `restFallbackTranscribe`, the fallback gets an empty buffer and silently returns null. Tied to #2 — the WAV salvage is only safe for strictly serial sessions.

### Important follow-ups

**4. Settings toggle is not live.**
`handleStreamingAudioToggle` (`settings.tsx:594-601`) only PUTs the server flag. `app.tsx` reads the flag once on mount (1089-1097) into a module-level var with no `onStreamingAudioChanged` IPC listener among the live-change subscriptions (1104-1115). Enabling has no effect until the pill window is recreated, and disabling never resets `_streamingAudioEnabled` for the session. Wire an IPC event like the existing `onOutputModeChanged`/`onAudioDuckingChanged`.

**5. AudioWorklet node connected to `ctx.destination`.**
`streamer.ts:97` connects the PCM worklet to `destination`. The worklet writes no output so it's likely silent, but connecting raw mic PCM to the speakers is a latent feedback risk and unnecessary — the analyser path in `app.tsx` deliberately never connects to destination. Drop the connection.

**6. Deepgram `cancel()` triggers an unwanted upstream reconnect.**
Deepgram is keep-warm, so cancel calls `upstream.cancel()` (`stream.ts:758`). `cancel()` sends `{type:"CloseStream"}` but never `ws.close()` (`deepgram.ts:169-180`); Deepgram then closes server-side → `onClose` → route reconnects (`stream.ts:572-580`) since `upstream===session` and `closed===false`. Wasteful reconnect + possible spurious `session.ready` after cancel.

**7. ElevenLabs token fetch has no timeout.**
`getSingleUseToken()` (`elevenlabs.ts:45-59`) uses no `AbortSignal.timeout`, unlike every other cloud call. A hung token endpoint means the session never opens and audio buffers up to the 500-chunk cap with no upstream timeout.

### Style / consistency

**8. New config calls bypass the typed `hc` client.** The renderer already uses `getClient()` (`hc<AppType>`) for sibling settings/usage/auth calls, and `/api/config` is part of the exported `AppType`. The new `/api/config` GET (`app.tsx:1089`) and `/api/config/flags/:key` PUT (`settings.tsx:596`) use raw `fetch` instead. Prefer `getClient()` here for consistency. (Note: the `/api/transcribe` calls correctly stay on raw `fetch` — they send a binary WAV Blob with custom `x-*` headers, which `hc` doesn't model well, and the new fallback at `app.tsx:376` intentionally mirrors the existing batch call.) Also worth typing the config route body with `zValidator` so the client gets request inference like the neighboring settings route.

Not merging — leaving as draft for these to be addressed.

MathurAditya724:
Marked ready. Pushed `c257314` addressing all 8 review findings:

- Streamer is now destroyed on unmount and on flag-off (was leaking a WebSocket reconnect loop + AudioContext for the app lifetime)
- Re-record is blocked while a streaming commit is pending, so the single WS/PCM buffer can't get its resolver overwritten or audio wiped mid-session (batch overlap is unaffected)
- `getWavBlob()` is non-destructive so the REST fallback survives an error+timeout race
- `settings:streaming-audio-changed` IPC makes the toggle live without recreating the pill window
- PCM worklet no longer connects to `ctx.destination` (feedback risk); verified against the Web Audio spec that source→worklet alone keeps `process()` pulled
- Deepgram `cancel()` keeps the warm socket open instead of sending `CloseStream`, which had forced a spurious reconnect
- ElevenLabs single-use token fetch now has a 10s timeout
- `/api/config` calls go through the typed `hc` client and the flag PUT body is validated with `zValidator`

CI is green on this commit: all 14 checks pass (Typecheck, Lint, Knip, Server API + Electron E2E tests, Build on Linux/macOS/Windows, server image, plugins). Locally: server `tsc` clean, electron `typecheck` clean, 209/209 server tests + 4/4 electron tests, biome clean.

A follow-up re-scan of the fixes surfaced one narrow, gracefully-degrading edge case worth noting for reviewers: toggling streaming **off** during a live transcription (~15s window) discards the recoverable WAV and the session resolves as a timeout instead of salvaging via REST. Low severity and strictly better than the pre-fix behavior; left unaddressed to keep the diff tight, happy to harden if preferred.

MathurAditya724:
Follow-up in `63c145f` addressing the two staff-review items:

**1. Re-record UX parity (the real judgment call).** The previous fix silently dropped a hotkey press if you pressed record while a streaming commit was still finalizing — batch mode honors that same press by starting a new dictation. Rather than leave that asymmetry, the press is now remembered (`pendingReRecordRef`) and replayed via `startRecording(true)` once the pending commit resolves, so streaming matches batch. The intent is cleared on `hidePill` and guarded by `!wantsMicRef.current` so a stale replay can't fire after cancel or if something else already took the mic.

**2. Tests for the two previously-untested server fixes:**
- `deepgram-cancel.test.ts` — asserts `cancel()` keeps the warm socket open (no `CloseStream`, no `close()`), and that `close()` still tears it down. Verified it fails if the old `CloseStream` behavior is reintroduced.
- `config-route.test.ts` — asserts the flag PUT round-trips a boolean and rejects non-boolean / missing values with 400 (the `zValidator` change).

Suite now 214 passing (was 209; +5 across the 2 new files). Server + electron typecheck clean, biome clean.

PR #473, opened 2026-07-16
feat(recording): add separate hold and toggle hotkeys

### LINKED ISSUE #452: [Feature] Have two separate hotkeys for hold mode vs toggle mode
# Overview

Right now, we have a single hotkey for both the hold mode and the toggle mode, and the user has to select whether or not they want to use each toggle in the app. We want to change this so that the hold mode has its own hotkey and the toggle mode has a separate hotkey, and they both cannot be the same. 

Image

Take a look at Wispr Flow's push to talk mode hands-free mode for inspiration: 

Image

# How to fix
1. Take a look at how we currently implement toggle mode versus hold mode, then figure out how to hotwire two separate hotkeys for each setting. 
2. Add the new hotkey setting into the settings page. Make sure that the hotkey for toggle mode cannot be the same as the hotkey for hold mode.

GatikTulsiani07:
## Summary

Implements #452 by replacing the single hotkey plus activation-mode model with two independent recording shortcuts:

- **Hold to record**
- **Toggle recording**

Both shortcuts can now be configured at the same time while coordinating through one shared recording session.

### Runtime

- Added independent native listeners for Hold and Toggle.
- Preserved `SETTINGS_KEYS.hotkey` as the existing Hold binding.
- Added optional `SETTINGS_KEYS.hotkeyToggle` with no default.
- Hold uses key-down/key-up semantics.
- Toggle starts and stops on successive key-down events and ignores key-up.
- Both bindings share one recording state, so only one recording session can be active.
- Added generation and listener-identity guards to ignore stale callbacks during re-registration.
- Re-registration releases an active recording exactly once.
- Ordinary hotkey replacement unregisters only hotkey-manager-owned fallbacks, preserving unrelated shortcuts.

### Failure handling

- Hold does not silently degrade to Toggle semantics when its native listener is unavailable.
- Toggle may fall back to Electron `globalShortcut`, where Toggle semantics remain correct.
- A failure in one binding does not disable the other.
- Factory-reset failure resumes the preserved bindings instead of leaving the manager permanently disabled.

### Validation and persistence

- Added canonical comparison for aliases and modifier ordering.
- Prevents Hold and Toggle from using the same logical accelerator.
- Binding updates are serialized across both shortcut kinds.
- Runtime registration failure rolls back both persisted settings and the previous runtime pair.
- Toggle can be cleared; Hold cannot be cleared.

### Settings UI

- Replaced the old Hold/Toggle activation selector with two independent shortcut rows.
- Added Set, Change, and Clear actions where appropriate.
- Toggle displays **Not set** when unconfigured.
- Added localized duplicate and save-error messages.
- Kept the previous displayed value until a save succeeds.

## Compatibility

No database migration is required.

- Existing `hotkey` values continue to represent Hold.
- Existing `hotkey_mode` rows are left untouched but are no longer used by the new flow.
- Legacy `hotkey:update`, `hotkey:reload`, and `hotkey:set-mode` compatibility paths remain available for existing consumers.

## Validation

Completed successfully:

- Electron typecheck
- Electron unit tests
- Electron build
- Native macOS helper compilation
- Biome
- `git diff --check`
- Focused Issue #452 Playwright coverage
- Runtime lifecycle, duplicate detection, rollback, serialization, recorder recovery, and settings UI tests

Full Electron E2E result:

- **55 passed**
- **1 known unrelated failure**: the existing sidebar test expects 7 navigation links while the current UI renders 9

## Manual verification

Verified on macOS:

- Both Hold and Toggle shortcuts can be configured independently.
- Both native listeners register successfully.
- Each shortcut starts the recording flow.
- Audio reaches the transcription endpoint.

The local transcription environment later reported that the `small-q5_1` Whisper model was not downloaded. That is existing local-model provisioning behavior and is unrelated to this hotkey change.

## Screenshot 
Screenshot 2026-07-16 at 3 39 11 PM

CI note: the Electron E2E job currently reports 55 passing tests and one unrelated existing failure in `tests/app.test.ts` (`sidebar navigation is rendered`). The assertion expects 7 sidebar links while the current UI renders 9. All Issue #452 hotkey tests pass.

Fixes #452


GatikTulsiani07:
Hey @MathurAditya724 @matteo8p I merged the latest main into this branch, resolved the merge conflict in settings.tsx, and reran the checks. At the moment, the only failing CI job is the existing Electron E2E test:

The workflow reports 55 passing tests and 1 failing test, and the failure appears to be the sidebar assertion rather than anything related to the changes in this PR.

I intentionally didn’t modify that test because it seems unrelated to Issue #452 and I wanted to keep this PR scoped to the independent Hold/Toggle hotkey work.

Would you prefer that I:

1. leave the PR as-is if this is a known unrelated failure, or
2. update the sidebar test in this PR as well?

Happy to do whichever you prefer. Thanks!

GatikTulsiani07:
I  have investigated the failing Electron E2E check. It turned out to be a stale sidebar test after the latest upstream navigation changes rather than an Issue #452 regression. I updated only the test to assert the current sidebar labels/routes and reran the Electron E2E suite (56/56 passing). No production code changed.

GatikTulsiani07:
All code checks are passing and the branch is now conflict-free.

The remaining Changelog Preview failure appears to be repository workflow-related: actions/checkout is refusing to check out fork PR code from the pull_request_target workflow.

Could a maintainer @matteo8p  please confirm whether this check is expected for fork-based PRs or needs a workflow-side adjustment?

PR #474, opened 2026-07-16
feat(plugins): forward beforeCleanup system fragments to Freestyle Cloud

MathurAditya724:
## Summary

Run the `beforeCleanup` plugin hook locally on all Freestyle Cloud paths (batch combined, post-process, and streaming) and forward the collected system-prompt fragments to the cloud in the same request. The cloud appends them to its assembled system prompt, so plugin instructions (e.g. emoji insertion) are honored without forcing a raw-mode round trip.

### Changes

**Plugin system fragment forwarding:**
- `freestyle-cloud.ts`: `transcribeWithFreestyleCloud()` and `postProcessWithFreestyleCloud()` accept `systemFragments?: string[]` and include them in the cloud request payload.
- `transcribe.ts`: On the batch combined cloud path, run `beforeCleanup` locally (with empty `input.text`), collect system fragments, and forward them via `transcribeWithFreestyleCloud()`.
- `post-process.ts`: Forward `promptHook.system` to `postProcessWithFreestyleCloud()` on the cloud cleanup path (where `input.text` is the real transcript).
- `stream.ts`: Run `beforeCleanup` at `connectUpstream()` time and include `systemFragments` in the streaming cleanup config sent to the cloud DO.
- `streaming/types.ts` + `streaming/providers/freestyle-cloud.ts`: Thread `systemFragments` through `StreamCleanupPreferences` and the `start` WebSocket message.

**Streaming session lifecycle:**
- `session-policy.ts`: `shouldKeepStreamingUpstreamAlive()` now returns `false` for `freestyle-cloud`. The upstream WebSocket closes after each transcription and reconnects on the next `start` message (hotkey-down), giving a natural pre-warm window.

**Dev-linked plugin asset fix:**
- `ui-assets.ts`: `resolvePluginAsset()` now follows the first symlinked segment when checking realpath containment, so dev-linked plugins with a `dist/` symlink pass the path-traversal guard.

**Emoji plugin:**
- New plugin at `plugins/emoji/` using the `beforeCleanup` hook to inject emoji instructions into the LLM cleanup prompt. Settings UI (placement mode: add-after vs replace) built with React + TanStack Query.

### Caveats
- On the batch combined and streaming cloud paths, `input.text` passed to `beforeCleanup` is empty (the transcript hasn't been produced yet). Plugins that contribute static fragments work fine; plugins that inspect the transcript won't get it on those paths.
- Requires companion cloud PR to accept and use the `systemFragments` field.

### Testing
- `tsc --noEmit` passes for both `tsconfig.node.json` and `tsconfig.web.json`
- Emoji plugin builds and links successfully


MathurAditya724:
Self-review found a correctness regression in the fragment-forwarding change and pushed a fix (d9231d0).

**Bug: `beforeCleanup` skip/abort/prompt-override silently ignored on the cloud cleanup paths.**

The early `beforeCleanup` run added on the combined-cloud (`transcribe.ts`) and streaming (`stream.ts`) paths only collected `system` fragments. It didn't act on the hook's other outputs, so:

- A plugin returning `skip: true` still got cloud-side LLM cleanup.
- A `consume()`/`abort()` in `beforeCleanup` still triggered a (wasted) cloud cleanup call.
- A `prompt` override was silently dropped.

This regressed the pre-forwarding behavior: previously `plugins().has("beforeCleanup")` forced the raw-STT + local post-process fallback, and `post-process.ts` honored `skip`/`consume`/`abort` (see `post-process-control-cloud.test.ts`). Moving the hook earlier without wiring those outcomes broke that contract.

**Fix:**
- `transcribe.ts`: when the hook skips, goes terminal, or overrides the prompt, drop out of combined mode (`mode: "raw"`) so the local post-process path applies the decision — matching the local/BYOK flow. `system`-only plugins (e.g. emoji) keep the fast combined path.
- `stream.ts`: streaming has no local cleanup path, so set `skipPostProcess` when the hook skips/aborts, and adjust `cloudHandledPostProcess` so the raw text flows through the cleanup-off branch (afterTranscribe/Transcribed still fire).
- Added `transcribe-cloud-combined-fragments.test.ts` covering fragment forwarding vs. skip/prompt/consume dropping to raw mode. This was the test-gap that let it through.

`tsc --noEmit` and the full server suite (262 tests) pass.

Note (not fixed, minor): the dev-link guard in `ui-assets.ts` only relaxes containment for a symlink on the *first* path segment (breaks at the shallowest existing segment), so a dev-link where a deeper segment is the symlink won't be served. Security posture is intact — the check still requires the real path to live under a realpath'd ancestor. Flagging for awareness only.

PR #475, opened 2026-07-16
feat(settings): connect to a self-hosted server from the Network tab

MathurAditya724:
## Summary

Restores the ability to point the desktop app at a **self-hosted Freestyle server** (with an optional bearer token), which was removed in #342. The option lives in **Settings → Network**, after the enterprise proxy/CA section, and it applies **without an app restart**.

This is the planned follow-up to #462's deferred "remote-server auth" note.

## No-restart design

- The local embedded server **always keeps running** — startup is never gated on the configured URL, so switching back to local (or between remotes) never needs a restart.
- On save: the settings window re-points its API client (`refreshApiBase()`) + refetches everything (`queryClient.invalidateQueries()`); the main process broadcasts `server:changed` so the pill window re-points and rebuilds its streamer. The pill also already calls `refreshApiBase()` before every dictation as a backstop.

## Storage

Server URL/token are stored in Electron's `settings.json` via IPC (`serverUrl`/`serverToken`) — **not** SQLite. They're client-side config (which server to talk to), so they can't live on the server itself.

## Server-side auth

Bearer-token middleware wired directly into `createApp()` (after `pluginApiGuard`). It's a **no-op unless a token is configured**, so the loopback Electron server is unaffected. Standalone/Docker servers read `FREESTYLE_AUTH_TOKEN` (via `startServer({ token })` / `startup.ts`). `/api/health` stays exempt; WebSocket upgrades authenticate via `?token=` (browsers can't set the `Authorization` header on a socket).

## Changes

- **packages/validations**: restore `serverUrlSchema` (empty = local server; validates + normalizes URL)
- **main / preload**: `server:url`/`token` IPC + `server:changed` broadcast; auth headers on `relayEvent` / `putServerSetting`
- **renderer `api.ts`**: `serverUrl`/`token` state, `getAuthHeaders`, `checkServerAuth`, auth-carrying `getClient()`
- **`settings.tsx`**: `ServerConnection` card (endpoint, token w/ reveal, Test / Save / Reset, status chip)
- **`streamer.ts`**: bearer token via `?token=` WS query param
- **server**: `authMiddleware` in `createApp`; `startServer({ token })`; `startup.ts` reads `FREESTYLE_AUTH_TOKEN`
- **tests**: `auth.test.ts`, `server-url.test.ts`

## Known limitation

Matches #462's own deferral: plugin **UI pages** over a *token-protected remote* server won't authenticate yet (a `WebContentsView` can't send the bearer header). Local server and non-token remotes are fully covered; all primary data paths (settings, history, transcribe, plugin management) carry auth.

## Verification

- Server tests: **273 passed** (incl. new `auth.test.ts` + `server-url.test.ts`)
- Server build (tsc): clean
- Electron typecheck (node + web): clean
- Biome + knip: clean (no new issues)


matteo8p:
I don't think we should ship this to production @MathurAditya724

PR #477, opened 2026-07-16
feat(plugins): voice-commands plugin (LLM agent action routing)

MathurAditya724:
## What

Adds the **`@freestyle-voice/commands`** plugin: turn spoken trigger phrases into actions — run macOS Shortcuts, call webhooks, open URLs/apps, or execute scripts by voice.

Say _"post to Slack that I'll be five minutes late"_ and, instead of typing the words, Freestyle runs the matching action.

## Supersedes #379

#379 introduced this plugin **plus** two SDK/server seams. The `#462` plugin-pipeline overhaul has since landed those capabilities in a better form, so this PR keeps only the plugin and rebuilds it against the current SDK. #379 will be closed in favor of this.

| #379 (old) | current SDK (this PR) |
|---|---|
| `afterTranscribe` `output.consumed = true` | `api.control.consume(reason)` |
| `PluginContext.llm` (captured in `setup`) | `api.llm` (built per-dictation, passed to the hook) |
| bridge `res.json<T>()` | native same-origin `Response` → `res.json()` |

Because the LLM now flows through `api.llm`, the agent also works for **Freestyle Cloud** users once #476 (the `freestyle-cloud` LLM provider) and freestyle-voice/cloud#65 (the `/v2/llm` endpoint) land.

## How it works

1. **Prefilter** — a cheap deterministic phrase match gates everything. No trigger → the transcript is dictated normally.
2. **Agent** — on a match, a multi-step tool-calling agent (`api.llm`) decides whether it's really a command and extracts the payload. Falls back to a deterministic longest-trigger match when no LLM is available.
3. **Action** — the command fires and the utterance is consumed via `api.control.consume()` (cleanup skipped, no text delivered).

### Action types
- **Run Shortcut** *(macOS only, hidden elsewhere)* — runs a macOS Shortcut by name; payload piped to stdin.
- **Call webhook** — `POST` payload as JSON, or `GET` as a query param.
- **Open URL / app** — opens a URL or app scheme; `{{input}}` substituted.
- **Run script** — runs a shell command; `{{input}}` substituted and payload also in `$FREESTYLE_COMMAND_INPUT`.

CRUD + a platform probe are exposed via plugin middleware; a React UI manages commands.

## Notes

- Bundles `ai` + `zod` in `devDependencies` (installed plugins get no transitive `npm install`); the host-created model is passed to the plugin's bundled `generateText` — both are AI SDK v6, so the model spec object is compatible across copies.
- Must still be added to the `plugins` allowlist (enabled in the UI) for the server to run it, like every plugin.
- Known follow-up: `ROUTE_BASE` / UI `ROUTE` hardcode the production slug; dev-linking under a `-dev` slug would 404 the settings page (documented gotcha). Not addressed here to keep this focused on the API migration.

## Verification

- `@freestyle-voice/commands` typecheck (server + UI) ✓
- full build (`pkgroll --minify && vite build`) ✓
- `biome check` clean ✓

### DIFF
--- a/plugins/voice-commands/README.md
+++ b/plugins/voice-commands/README.md
@@ -1,0 +1,31 @@
+# Voice Commands
+
+Turn spoken trigger phrases into actions. Define commands like _"post to Slack…"_
+or _"add a reminder…"_ and Voice Commands will detect them in your dictation and
+run the matching action instead of typing the text.
+
+## How it works
+
+1. **Prefilter** — a cheap, deterministic phrase match gates everything. If your
+   speech contains no trigger phrase, nothing else runs and the transcript is
+   dictated normally.
+2. **Agent** — when a trigger matches, a multi-step tool-calling agent (using the
+   LLM you've already configured for cleanup) decides whether it's really a
+   command and extracts the payload (the part of your speech that isn't the
+   trigger).
+3. **Action** — the command fires. The utterance is _consumed_: cleanup is
+   skipped and no text is delivered to the focused app.
+
+If no cleanup LLM is configured, detection falls back to a deterministic match
+(the longest matching trigger wins) so commands still work.
+
+## Action types
+
+- **Run Shortcut** _(macOS only)_ — run a macOS Shortcut by name; the payload is
+  piped to its input.
+- **Call webhook** — `POST` the payload as JSON, or `GET` it as a query param.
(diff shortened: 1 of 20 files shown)


MathurAditya724:
Closing this in favour of #481

PR #495, opened 2026-07-25
refactor: cleanup

MathurAditya724:
## Summary

Primary fix: **first-dictation-after-restart recording failure**, plus in-progress cleanup across several renderer pages, plus a new **Advanced mode** toggle that gates the Models page and a rename of the **Today** tab to **Transcriptions**.

### First-dictation race fix

The per-session transport decision (`recordingSessionUsesTransportRef`) was being mutated mid-recording by a late-arriving WebSocket `config` message in `onConfig`.

**Repro:** restart the Electron app, then do the first dictation → "No audio captured / Recording Failed". Subsequent dictations work.

**Root cause:**
1. On startup, `getStreamer()` opens the WS but `supportsSessionTransportRef` starts `false` (config not yet received).
2. First recording latches the **batch** capture path in `startRecording` (`app.tsx:879-887`) and starts a MediaRecorder — the streamer never captures audio.
3. The `config` message arrives mid-recording; `onConfig` flipped both `supportsSessionTransportRef` **and** (because `wantsMicRef` was true) `recordingSessionUsesTransportRef` to `true`.
4. On commit, `commitRecording` (`app.tsx:995`) now reads `true` → takes the **streaming** path, but `pcmSampleCount === 0` → empty commit → REST fallback with empty WAV → failure.

**Fix:** `onConfig` now updates only `supportsSessionTransportRef` (for *future* sessions). The per-session decision is latched exactly once in `startRecording` and never mutated mid-session, so capture and commit always agree.

### Rename "Today" tab → "Transcriptions"

The first sidebar tab is renamed from **Today** to **Transcriptions** to better describe its content. Updated across all 8 locale files (`en`, `template`, `de`, `es`, `fr`, `it`, `ja`, `pt`). Only the `shell.nav.today` label changed — the route (`/today`), component, and other `today.*` strings are untouched.

### Advanced mode toggle for the Models page

Gates the Models page behind an **Advanced mode** setting so the default experience stays simple.

- **Off by default for everyone** (new and existing users). The Models page is hidden from the sidebar; Freestyle Transcribe is used automatically.
- Opt-in via **Settings › Application** → *Advanced mode*. Flipping it patches the shared settings query cache so the sidebar updates instantly.
- Toggling off is **non-destructive** — configured models/API keys are kept and transcription keeps using them; only the sidebar tab is hidden.
- Models nav item is hidden when off, and `Cmd+N` shortcuts are re-numbered sequentially so there's no gap (e.g. Plugins becomes `Cmd+5`).
- Direct/deep-link access to `/settings/models` while off still renders the page, with an info banner pointing to the toggle.
- Adds the `advanced_mode` setting key and `en`/`template` strings.

> Note: this intentionally does **not** ship a schema migration. Existing users also start with Advanced mode off and can opt in whenever they want the Models page back — their configured models are preserved either way.

### Other cleanup (in-progress working-tree changes)
- `history.tsx` rewrite; `today.tsx` removed
- `shell.tsx` nav
- `dictionary.tsx`, `cloud-profile.tsx`, `dashboard.tsx`
- locales (`en.json`, `template.json`)
- `packages/validations/src/settings.ts`
- `apps/electron/src/main/index.ts`

## Test plan
- [x] `typecheck:node` + `typecheck:web` pass
- [x] Server typecheck + tests pass (294 tests)
- [x] Biome check clean on all changed files
- [x] All locale JSON files parse
- [ ] Restart Electron → first dictation transcribes (no "Recording Failed")
- [ ] Subsequent dictations still work
- [ ] Sidebar shows "Transcriptions" as the first tab
- [ ] Fresh + existing installs: Models tab hidden by default; dictation works
- [ ] Toggle Advanced mode on/off → Models tab appears/disappears immediately
- [ ] Visit `/settings/models` with Advanced mode off → banner shows, "Open settings" jumps to Application


matteo8p (approved):
Love the direction, and how clean the new UI looks. Also really love having the new advanced mode. Where we're hiding the models page. 

One thing to call out is to make sure that the advanced mode toggle is off by default. I forgot to check what the default was when I first opened up the app. But just do a sanity check and make sure that the advanced toggle is off by default.

MathurAditya724:
It's in Settings > Application. For current users it will be on by default but for new users it will be off. Should I change that?

matteo8p:
> It's in Settings > Application. For current users it will be on by default but for new users it will be off. Should I change that?

I am mostly indifferent. I think I might just be simple to keep it off by default for everyone. We don't really have a lot of current users to where we need to have separate logic for current users versus new users.

matteo8p (changes requested):
One small net is to rename the "Today" tab to something else, like “transcription” or something.

PR #509, opened 2026-07-27
feat: Voice pill redesign

tigerisaac:
## Summary

Cleaned up the voice pill and made it nicer to look at. Adds a new feature, a close button that you can click as an alternative to esc. In settings you can switch between always-present and hover-only.

## Checklist

- [X] PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (`type(scope): summary`)
- [X] Ran `pnpm biome check .` (lint + formatting)
- [X] Ran `pnpm --filter @freestyle-voice/electron typecheck:web` (if touching the app)


tigerisaac:
Mostly finished. Just touching up some stuff and will be back later.

MathurAditya724 (commented):
Reviewed the full diff, ran the new streamer test (passes), `typecheck:web` (clean once `@freestyle-voice/stt` is built — the errors otherwise are just monorepo build-order, not this PR), and `biome check` on all touched files (clean).

This is a careful change. I traced the paths I was most worried about and they hold up:

- **Draw loop** (`runBars`): no OOB (`barLinesRef.length === BARS === bars.length`), no div-by-zero, heights bounded, buffers reset on `startBarAnimation`/`stopVisualization`, and the "keep the rAF chain alive while the analyser is torn down" early-return re-schedules correctly. The old `setInterval` path was fully removed.
- **rAF lifecycle**: `startBarAnimation` cancels the prior frame; `stopVisualization` cancels + nulls; `hidePill` calls it. No orphaned loop.
- **Expand/collapse**: renderer schedules `setPillExpanded(false)` on `showCard→false`, main `hidePill()` also calls it, and `setPillExpanded` is idempotent (`expanded === isExpanded` early return) — the window can't get stuck grown. The slot-coordinate handling in the `move` listener (`rawX + pillExpandOffset.dx`) correctly stops the custom position from walking across expand/collapse cycles.
- **Streamer singleton closures** (`getStreamer`, empty deps): every captured callback is a stable `useCallback` reading refs, so they're not stale.
- **REST fallback is non-destructive** (`getWavBlob()` survives `cancel()`), and streaming-vs-fallback is guarded by nulling `streamResolverRef`, so a transcription can't be delivered twice.
- **Cancel-mode wiring**: state + ref + load-from-settings + IPC listener with cleanup + ref-sync effect all present; the `pointerEvents: open > 0.5` gate matches the "ignore until half open" note; `hidePill` and `handlePillLeave` both respect `"always"`.
- Dep removal (`three`, `@react-three/*`, `@types/three`) is clean — no leftover source references, `orb.tsx` / `perlin-noise.png` deleted.

One minor thing worth tightening:

**`app.tsx:1917-1920` — inner `requestAnimationFrame` isn't cancellable.** The cleanup returns `cancelAnimationFrame(frame)`, but `frame` is only the outer rAF; the inner rAF's id is never captured. If `showCard` flips to `false` (dismiss/retry) or the component unmounts in the ~16ms between the two frames, the inner callback still runs `setRoomReady(true)` — a stray state write (and a setState-after-unmount warning on unmount). Low impact since `cardOpen` also gates on `showCard`, but easy to close:

```ts
let inner = 0;
const outer = requestAnimationFrame(() => {
  inner = requestAnimationFrame(() => setRoomReady(true));
});
return () => {
  cancelAnimationFrame(outer);
  cancelAnimationFrame(inner);
};
```

Nothing blocking.

PR #516, opened 2026-07-29
fix(electron): improve permission handling for dictation startup

### LINKED ISSUE #511: [Bug] Ensure that microphone and accessibility permissions are always on; otherwise, throw an error.
# Overview 

Right now, if the user doesn't have the right accessibility permissions or microphone permissions, the app silently fails. Take a look at this screenshot: where a user didn't have accessibility permissions, but it was silently failing to paste. 

Image

OB1 is that if the user doesn't have the right microphone or accessibility permissions, we want to throw an error, and encourage them to set up the right permissions 

# How to fix
1. Start off by replicating the issue. Remove your own accessibility features and try to dictate without accessibility permissions to reproduce the issue. 
2. Go into the code and throw an error if the user tries to dictate, but accessibility is disabled. We should always flag an error, too, when the user starts the app but doesn't have accessibility permissions.

GatikTulsiani07:
## Summary

This PR improves the Electron permission flow by preventing dictation from starting when required permissions are unavailable and by providing clearer startup guidance for existing users.

### Fail-fast permission preflight

- Added a permission preflight before dictation starts.
- Prevented `RecordingStarted`, microphone capture, transcription, and paste from running when required permissions are unavailable.
- Preserved the existing recording flow when permissions are granted.

### Accessibility

- Use the live macOS Accessibility trust state instead of relying on stale cached state.
- Clear stale Accessibility permission state when trust has been revoked.
- Show startup warnings for existing users when Accessibility permission is required.
- Continue suppressing startup warnings during onboarding to avoid duplicate permission dialogs.

### Microphone

- Added startup warnings for denied and restricted microphone permissions.
- Continue using the existing `getUserMedia` flow when permission is `not-determined`.
- Show a combined startup warning when both Accessibility and Microphone permissions are unavailable.
- Preserve the existing Linux permission flow.

### Tests

- Added regression coverage for startup permission decisions.
- Added startup warning coverage.
- Added permission preflight tests.
- Added combined permission warning tests.
- Improved Electron E2E fixture isolation so onboarding state is deterministic and independent of existing user data.

## Validation

-  Biome
-  Node typecheck
-  Web typecheck
-  Electron tests
-  Electron production build
-  Full Electron E2E suite (54/54 passing)

Closes #511

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -398,6 +398,11 @@ jobs:
           APPLE_APP_SPECIFIC_PASSWORD: ${{ secrets.APPLE_APP_SPECIFIC_PASSWORD }}
           APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
 
+      - name: Run macOS permission flow E2E tests
+        working-directory: apps/electron
+        timeout-minutes: 5
+        run: pnpm run test:e2e tests/permission-flow.test.ts
+
       - name: Verify native binaries in package
         working-directory: apps/electron
         run: node scripts/verify-native-binaries.mjs
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -74,6 +74,7 @@ import {
   protocol,
   screen,
   shell,
+  systemPreferences,
   Tray,
 } from "electron";
 import { autoUpdater } from "electron-updater";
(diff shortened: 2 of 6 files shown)


GatikTulsiani07:
hey @matteo8p have a look at this PR

matteo8p (changes requested):
This seems to be a specific for macOS. Do you know how the permission checking can be done? On Windows and Linux?

GatikTulsiani07:
@matteo8p  This implementation only adds permission handling for the platform behaviour that I was able to validate directly. I implemented and tested the changes on macOS, where I verified the permission flow end-to-end. I don't currently have access to Windows or Linux machines, so I wasn't able to validate how the same flow should be implemented there or whether this project already has a preferred platform-specific approach.

Rather than introducing behaviour that I couldn't verify, I limited this PR to the implementation I could test. If there is an expected Windows/Linux permission flow or a preferred cross-platform approach for this repository, I'd be happy to update the implementation accordingly.

matteo8p (approved):

PR #518, opened 2026-07-29
feat(mobile): live voice-keyboard setup detection

MathurAditya724:
## What

Adds **live detection** of whether the Freestyle voice keyboard is enabled with Full Access on iOS, and surfaces guided setup in both onboarding and the keyboard-setup screen.

## Why

The mobile onboarding/settings previously only showed static text ("set up the voice keyboard later in Settings") with no feedback on whether the user actually completed the steps. iOS provides **no API** to programmatically enable a custom keyboard or grant Full Access (the same constraint Wispr Flow and every keyboard app lives under), so the best we can do is guide the user to Settings and reflect the result live.

## How — the detection insight

A keyboard extension can **only** write to the shared App Group `UserDefaults` when it has **Full Access**. So the extension now stamps a handshake timestamp on load; a non-zero timestamp read back by the app is honest proof the keyboard is **both enabled and has Full Access**.

- `markKeyboardActive()` / `keyboardLastActive()` added to the shared `DictationBridge` (both verbatim copies); stamped in the keyboard's `viewDidLoad`.
- Exposed `keyboardLastActive()` through the native module → TS declaration → `index.ts` → the platform-safe `dictation-bridge` facade (returns `null` on Android / Expo Go).
- New `useKeyboardStatus` hook: `"unsupported" | "not-enabled" | "ready"`, re-reads on app foreground + a 1.5s poll so it flips the moment the user returns from Settings.
- `keyboard-setup.tsx`: live status banner, steps check off, "Open Settings" CTA hides once ready.
- `onboarding.tsx` step 0: live "Voice keyboard" row (iOS only) with a "Set up" → Settings deep link that flips to a checkmark when detected. Mic grant still gates Continue.

## Testing

- `pnpm --filter @freestyle-voice/mobile typecheck` ✅
- `pnpm --filter @freestyle-voice/mobile lint` ✅
- Native handshake requires a device/dev build to verify end-to-end (Expo Go can't run the keyboard extension); the JS facade returns `unsupported` there and the UI degrades gracefully.

## Notes

- Native IAP billing (replacing the Stripe web checkout) is intentionally **out of scope** and will follow in a separate PR.


MathurAditya724 (commented):
Reviewed the full diff. Core mechanism is sound — using the App Group `UserDefaults` write as proof-of-Full-Access is the right call given iOS exposes no API to query keyboard/Full-Access state. Typecheck + lint pass, the layering is clean, and the two `DictationBridge.swift` copies are byte-identical.

A few notes, none blocking:

**1. `"ready"` is sticky — revoking Full Access never flips it back** (`use-keyboard-status.ts` `readStatus`)
Status is `"ready"` for any `lastActive > 0`. The timestamp is only ever written (on keyboard load), never cleared or compared to a freshness window. Grant Full Access once, later disable it → the app reports "ready" forever. Partly inherent (iOS gives no revocation callback), but the docstring oversells it: the hook header says "a **recent** (or ever-present) handshake is honest proof" and frames the poll as freshness-driven, yet recency is never actually checked. Either drop "recent" from the comment or add a staleness window. Note a pure TTL would false-negative an enabled-but-unused keyboard (value only updates on load), so it's a real tradeoff — but the comment should match the code.

**2. `.lore.md` churn** — the diff pulls in ~20 entries from other merged PRs (account unlink, cloud OAuth, org endpoints, dropdown fixes, profile redesign) and adds zero documenting this feature. It's an auto-synced managed file so this is expected noise, but the Full-Access handshake trick is exactly the kind of non-obvious insight worth capturing as its own lore entry.

**3. Minor** — `useKeyboardStatus` returns `refresh`, unused by both consumers (foreground + poll auto-refresh covers current call sites). Dead surface for now.

Nice work — the detection insight is clever and the graceful `unsupported` degradation on Android/Expo Go is handled correctly throughout.

MathurAditya724:
Hey @MathurAditya724 — thanks for the thorough review. Addressed all three notes in 088d119:

1. **Sticky `"ready"` / docstring mismatch** — rewrote the `useKeyboardStatus` header to match the code: the signal is *deliberately* sticky (write-only, never freshness-checked), so revoking Full Access isn't detected, and I documented why a TTL can't be used (value only updates on keyboard load → an enabled-but-unused keyboard would false-negative). Dropped the misleading "recent"/freshness framing.
2. **Unused `refresh`** — removed from `KeyboardStatusResult`; foreground + poll auto-refresh covers every call site.
3. **Lore** — captured the App-Group-write-as-proof-of-Full-Access trick as its own gotcha entry.

Typecheck + lint still green.

PR #523, opened 2026-07-30
feat: Create the ability to translate on the fly

matteo8p:
## Summary
In this PR we introduce the ability to translate on the fly. In the transcription settings, language now has a translate mode. When it's on translate mode, it enforces that language. 

For example, if you have Spanish selected, even if you speak in English, it will translate to Spanish in real time.


apps/electron/src/renderer/src/pages/settings.tsx
```diff
@@ -169,7 +177,7 @@ export default function SettingsPage(): React.JSX.Element {
   const [autoUpdate, setAutoUpdate] = useState(true);
   const [launchAtStartup, setLaunchAtStartup] = useState(false);
   const [showOnLaunch, setShowOnLaunch] = useState(true);
-  const [advancedMode, setAdvancedMode] = useState(false);
+  // const [advancedMode, setAdvancedMode] = useState(false);
```
MathurAditya724: why are we doing this?

matteo8p: I just brought back advanced mode.

PR #527, opened 2026-08-01
fix(electron): default Windows push-to-talk hotkey to Right Alt

andstonk:
Windows previously defaulted to the 3-key Control+Alt+Space chord. Match macOS's single dedicated-key pattern (Fn) by defaulting to Right Alt instead. Also extends the settings hotkey recorder's DOM fallback capture to distinguish left/right modifiers on Windows and Linux (it only ever handled macOS), which is now required since a lone right-side modifier can be a valid recorded hotkey.

## Summary

- Windows now defaults to the single Right Alt key for push-to-talk, matching macOS's single dedicated-key pattern (Fn), instead of the 3-key Control+Alt+Space chord
- Extends the Settings hotkey recorder's DOM-capture fallback to distinguish left/right modifiers on Windows and Linux — previously only macOS could tell AltRight from AltLeft, which silently collapsed a recorded Right Alt into generic Alt

## Checklist

- [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (`type(scope): summary`)
- [x] Ran `pnpm biome check .` (lint + formatting)
- [x] Ran `pnpm --filter @freestyle-voice/electron typecheck:web` (if touching the app)

### DIFF
--- a/apps/electron/src/renderer/src/hooks/use-hotkey-recorder.ts
+++ b/apps/electron/src/renderer/src/hooks/use-hotkey-recorder.ts
@@ -248,14 +248,13 @@ export interface RightModifierUpdateEvent {
   explicitLeft?: boolean;
 }
 
-/** Maps DOM `KeyboardEvent.code` to the side-specific token it represents (macOS only). */
+/** Maps DOM `KeyboardEvent.code` to the side-specific token it represents. */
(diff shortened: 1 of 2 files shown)


matteo8p (approved):
Ship it

PR #531, opened 2026-08-01
feat: Show the update banner everywhere

matteo8p:
## Summary

In this PR, we show the update to latest version banner everywhere in the app. It stays persistent until the user updates the electron app.


MathurAditya724:
Thank you for taking care of this, really wanted this

PR #540, opened 2026-08-03
fix(pill): better handling of short transcriptions

tigerisaac:
This PR makes short dictations keep text.
Audio dropped during connect before, which is now fixed. 
There's no actual change to the voice pill (I initially planned to add a waiting set before the voice pill, but I found that just keeping the audio during initialization is enough)


apps/server/src/lib/streaming/providers/deepgram.ts
```diff
@@ -146,10 +149,15 @@ export class DeepgramTranscriptionProvider implements TranscriptionProvider {
 
     return {
       sendAudio(chunk: ArrayBuffer): void {
+        if (ws.readyState === WebSocket.CONNECTING) {
```
MathurAditya724, line 163: Audio is now held while the socket is `CONNECTING`, but `commit()` just below still calls `deliverFinal()` immediately when the socket isn't `OPEN`, resolving an empty transcript (`accumulatedText`/`partialText` are still empty) and discarding the audio that was just held. Soniox handles this correctly — its `commit()` returns on `CONNECTING` and finalizes in the `open` handler after the pending flush.

In the common route path this is masked because `stream.ts` defers the commit via `pendingCommit` until `onReady`. But Deepgram is kept warm and, on a keep-warm reset while a reconnect is still in flight, the route marks the reset session ready synchronously (`notifySessionReady`) and then forwards `commit()` while the socket is still `CONNECTING` — so a short dictation can still resolve empty even though the audio is held. Consider deferring the finalize like Soniox for full symmetry.

apps/server/src/lib/streaming/providers/openai.ts
```diff
@@ -149,14 +165,12 @@ export class OpenAITranscriptionProvider implements TranscriptionProvider {
 
     return {
       sendAudio(chunk: ArrayBuffer): void {
+        if (ws.readyState === WebSocket.CONNECTING || !configured) {
+          pending.hold(chunk);
```
MathurAditya724, line 187: Same asymmetry as Deepgram/ElevenLabs: `commit()` calls `deliverFinal(partialText)` when the socket isn't `OPEN`, so a commit that lands before `session.updated` resolves empty even though the audio is now held and flushed afterward (by which point `finalDelivered` is already true and no `input_audio_buffer.commit` is ever sent). The route's `pendingCommit` covers the common path, but the provider itself doesn't — worth aligning with Soniox's connecting-safe `commit()`.

apps/server/src/lib/streaming/pending-audio.ts
```diff
@@ -0,0 +1,46 @@
...
+  /** Drop what's held — a cancelled or restarted session. */
+  clear(): void;
+}
+
+export function createPendingAudio(): PendingAudio {
+  let chunks: ArrayBuffer[] = [];
+
+  return {
+    hold(chunk: ArrayBuffer): void {
+      if (chunks.length >= MAX_PENDING_CHUNKS) return;
```
MathurAditya724: Nit: at the cap this drops the *newest* chunk (the tail of the take) rather than the oldest. For the documented "session that never opens" case it doesn't matter, but for a session that stalls and then does open you'd silently lose the end of the dictation rather than the start. Given the cap is ~1 min of audio this is unlikely to bite in practice — just flagging the drop order since the buffer is explicitly ordered oldest-first.

apps/server/tests/short-dictation.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    const { session, socket } = openSession("openai");
+
+    shortDictation(session);
+    socket.open();
+    // Open is not enough — OpenAI rejects audio before session.updated.
+    expect(socket.audio).toHaveLength(0);
+
+    socket.reply({ type: "session.updated" });
+
+    expect(socket.audio).toHaveLength(2);
```
MathurAditya724: This asserts the held audio reaches the wire, but not that `onFinal` returns it. Since `shortDictation()` calls `commit()` while the socket is still `CONNECTING`, OpenAI's `commit()` takes the not-`OPEN` branch and `deliverFinal(partialText)` fires with empty text at commit time — the flush here happens afterward with `finalDelivered` already true. An `expect(onFinal).toHaveBeenCalledWith(...)` assertion (like the Soniox tests) would surface that gap.

Also: Deepgram and ElevenLabs gained the same connecting-hold behavior in this PR but aren't exercised here — might be worth folding them into the parametrized `openSession` coverage.

MathurAditya724 (commented):
Reviewed the streaming changes. The consolidation into `createPendingAudio()` is clean, and the Soniox `maybeDeliverAfterFinalize` empty-guard is a solid, well-explained fix that the new tests cover nicely.

One cross-cutting thing worth a look: the audio *hold* is now uniform across providers, but the `commit()`-before-open handling isn't. Soniox defers finalize until the socket opens, while Deepgram / OpenAI / ElevenLabs still `deliverFinal()` immediately when the socket isn't `OPEN` — which resolves an empty transcript and discards the audio that was just held. In the normal route path this is masked by `stream.ts`'s `pendingCommit` deferral, so it's narrow (the keep-warm reconnect-in-flight window), but it does mean the newly-held audio is wasted for those providers in exactly the short-dictation case this PR targets. Details inline. Nothing blocking.

apps/server/src/lib/streaming/providers/deepgram.ts
tigerisaac: Fixed in 7ef81e5: Deepgram now keeps a pre-open commit pending, flushes the held audio after the socket opens, and sends Finalize afterward. The timeout and reset/cancel state are covered by the provider regression tests.

apps/server/src/lib/streaming/providers/openai.ts
tigerisaac: Fixed in 7ef81e5: OpenAI now defers the commit until session.updated, flushes held audio first, and sends input_audio_buffer.commit exactly once. ElevenLabs received the same connecting-safe handling, and the short-dictation tests cover both paths.

apps/server/src/lib/streaming/pending-audio.ts
tigerisaac: Fixed in 7ef81e5: reaching the pending-audio cap now evicts the oldest chunk so the newest audio is retained. Added a regression test for the eviction order.

apps/server/tests/short-dictation.test.ts
tigerisaac: Fixed in 7ef81e5: short-dictation coverage is parametrized across Soniox, Deepgram, ElevenLabs, OpenAI, and Freestyle Cloud, with final-callback assertions for the providers that finalize through the upstream socket. All 402 server tests pass.

PR #541, opened 2026-08-03
feat(pill): give the voice pill an arrival and some better endings

tigerisaac:
Four beats now: arrival, listening, handover, departure.
This branch and its commits are to make the pill more animated and give it some small details that make it better to use. 
- Arrival: the enter transition existed but never ran — React mounted the subtree with `data-show="true"` already set, and CSS transitions don't fire on first paint. Held for one committed frame, then released; the bars stagger in right-to-left, the direction samples travel through the row.
- Handover: commit used to zero the targets and swap both easing constants in one frame, deleting your last syllable. It now settles to rest over 180ms, holds still for a beat, and only then starts the machine's sweep.
- Departure: delivered, cancelled and quiet were one instant `window.hide()`. Delivered closes the row from the outside in — every bar collapsing in place, so no two round caps ever slide into each other — draws a check out of the centre in the same white, and holds it. Cancelled gets half the entrance and no mark.
- Silence: 1.6s under the noise floor and the row admits it, rather than letting a muted mic look exactly like a quiet room until the "No audio captured" dialog.
- The capsule gains a soft shadow (it had none, and dissolved into busy backdrops) and lifts under the pointer, so `cursor: grab` has a partner.

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -1285,10 +1285,20 @@ async function deliverOutput(
 
   try {
     if (mode === OutputMode.Paste) {
-      await pasteIntoFocusedApp(text, async () => {
-        hidePill();
-        await wait(0);
-      });
+      // The pill deliberately stays up through the paste. It used to be hidden
+      // here, immediately before the synthetic keystroke, which made the
+      // renderer's delivered animation unobservable: by the time `pasteText`
+      // resolved and the capsule began closing into its confirmation mark,
+      // the window it was drawing into had been gone for hundreds of ms.
+      //
+      // Nothing about the paste needs the window down. The pill is created
+      // `focusable: false` (plus a non-activating `panel` on macOS), so it is
+      // never the focused window and the keystroke goes to the same app either
+      // way. Hiding is the renderer's job now — it owns the end of the
+      // session, and calls `pill:hide` when its exit finishes. Both callers of
+      // this function are renderer-driven, so there is always something on the
+      // other side to do it.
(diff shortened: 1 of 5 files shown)


apps/electron/src/main/index.ts
```diff
@@ -1285,10 +1285,20 @@ async function deliverOutput(
 
   try {
     if (mode === OutputMode.Paste) {
-      await pasteIntoFocusedApp(text, async () => {
-        hidePill();
-        await wait(0);
-      });
+      // The pill deliberately stays up through the paste. It used to be hidden
```
matteo8p: @tigerisaac Can you remove the code comments?

apps/electron/src/renderer/src/lib/streamer.ts
```diff
@@ -361,8 +361,10 @@ export class Streamer {
           this.flushPendingChunks();
           this.callbacks.onReady();
           break;
+        // Live partials are deliberately not surfaced anywhere in the UI —
```
matteo8p: Remove code comments

matteo8p (changes requested):

apps/electron/src/main/index.ts
tigerisaac: got it

apps/electron/src/renderer/src/lib/streamer.ts
tigerisaac: got it
Opus 5 yaps a lot. shoulda cought that

MathurAditya724 (approved):

PR #561, opened 2026-08-08
fix(electron): keep Right Alt distinct from Alt when recording hotkeys

### LINKED ISSUE #560: Windows: Right Alt cannot be re-recorded as the dictation hotkey (saves plain Alt)
## Description

On Windows the default dictation (push-to-talk) hotkey is **Right Alt**. If you rebind it to some other key and then try to rebind it *back* to Right Alt, the recorder never verifies/accepts Right Alt — it saves plain `Alt` instead, which then fires on **both** Alt keys. There is no way to get back to the shipped default from the UI.

## Steps to reproduce

1. On Windows, open Settings → Hotkeys. The dictation hotkey is `Right Alt` (the default).
2. Record a different hotkey (e.g. `F9`) and save.
3. Start recording again and press **Right Alt**.
4. Look at the saved hotkey.

## Expected

The recorder captures `Right Alt` and saves `RightAlt` — same as the shipped default. Left Alt does not trigger dictation.

## Actual

It saves plain `Alt`. Dictation now triggers on **both** Alt keys, and the Right Alt default cannot be restored by re-recording.

## Environment

- OS: Windows 11
- App version: 0.7.1

## Root cause (investigated)

In `apps/electron/src/renderer/src/hooks/use-hotkey-recorder.ts`, the two native binaries report right-side modifiers through different channels:

- **macOS** emits `RIGHT_MOD_DOWN:RightOption`, which arrives as a *modifiers update* and runs through `nextRightModifierLatch`, so the side is tracked.
- **Windows** emits `RECORD_KEY:RightAlt`, which main wraps as a *captured combo* — and the `onHotkeyRecordCaptured` handler cleared the right-modifier latch unconditionally.

`normalizeCapturedCombo` then collapses `RightAlt` to the generic `Alt` in the draft (intentional, so chords like `Right Alt + U` still work), but the latch it just cleared was the only thing that would have restored `RightAlt` on save. So recording Right Alt on Windows always persisted plain `Alt`.

The native C binary is not at fault — it distinguishes `VK_LMENU` from `VK_RMENU` correctly; the wrong accelerator simply never reached it.

Because Windows already *defaults* to `RightAlt`, this only surfaces after rebinding: the shipped default works, but re-recording Right Alt cannot reproduce it.

**Right Ctrl, Right Shift and Right Super on Windows have the identical problem.**

Note: fixing the recorder only corrects what recording *writes* — an already-saved `Alt` stays until the hotkey is re-recorded once.

andstonk:
## Summary

Fixes #560 

On Windows, recording **Right Alt** as the dictation hotkey saved plain `Alt`, which fires on both Alt keys. Since Windows already *defaults* to `RightAlt`, this only surfaced after rebinding — the shipped default worked, but re-recording Right Alt could not reproduce it, so there was no way back to the default.

**Cause.** The two native binaries report right-side modifiers through different channels, and the recorder only tracked the side on one of them:

- macOS emits `RIGHT_MOD_DOWN:RightOption`, which arrives as a *modifiers update* and runs through `nextRightModifierLatch`.
- Windows emits `RECORD_KEY:RightAlt`, which main wraps as a *captured combo* instead — and the captured handler cleared the right-modifier latch unconditionally.

`normalizeCapturedCombo` then collapses `RightAlt` to the generic `Alt` in the draft (intentional, so chords like `Right Alt + U` still work), but the latch it just cleared was the only thing that would have restored `RightAlt` on save. The C binary was never at fault — it distinguishes `VK_LMENU` from `VK_RMENU` correctly; the wrong accelerator simply never reached it.

**Fix.** Extract `latchFromCapturedCombo` — a captured key that is itself a side-specific modifier now *sets* the latch rather than clearing it. Ordinary keys still clear it, and a right modifier captured alongside other held modifiers stays a chord and does not latch. Right Ctrl, Right Shift and Right Super on Windows had the identical problem and are fixed by the same change.

With the settings window focused, the DOM keydown and the native IPC event both fire for one press in nondeterministic order, so the latch has to be idempotent either way: DOM-first sets it and the native capture confirms it; native-first sets it and the DOM path preserves it via the `rightToken === previousLatch` branch. Both interleavings are covered by the new tests.

**Note:** this corrects only what recording *writes*, so an already-saved `Alt` stays until the hotkey is re-recorded once.

## Testing

- New test cases in `use-hotkey-recorder.test.ts` covering solo Right Alt capture, the chord case, and both DOM/native interleavings — `pnpm --filter @freestyle-voice/electron test` → 11 passed.
- `pnpm biome check` on the touched files → clean. (A repo-wide run also reports pre-existing formatting diagnostics in `biome.json`, `package.json`, `turbo.json` and `skills-lock.json` that are unrelated to this branch and untouched here.)
- `pnpm --filter @freestyle-voice/electron typecheck:web` → clean.

## Checklist

- [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (`type(scope): summary`)
- [x] Ran `pnpm biome check .` (lint + formatting)
- [x] Ran `pnpm --filter @freestyle-voice/electron typecheck:web` (if touching the app)

### DIFF
--- a/apps/electron/src/renderer/src/hooks/use-hotkey-recorder.test.ts
+++ b/apps/electron/src/renderer/src/hooks/use-hotkey-recorder.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from "vitest";
+import {
+  type HotkeyCombo,
+  latchFromCapturedCombo,
+  nextRightModifierLatch,
+} from "./use-hotkey-recorder";
+
+const EMPTY: HotkeyCombo = { modifiers: [], key: null };
+
+describe("latchFromCapturedCombo", () => {
+  it("latches a solo right-side modifier", () => {
+    expect(latchFromCapturedCombo({ modifiers: [], key: "RightAlt" })).toBe(
+      "RightAlt",
+    );
+    expect(latchFromCapturedCombo({ modifiers: [], key: "RightControl" })).toBe(
+      "RightControl",
+    );
+  });
+
+  it("does not latch a right-side modifier held as part of a chord", () => {
+    expect(
+      latchFromCapturedCombo({ modifiers: ["Control"], key: "RightAlt" }),
+    ).toBeNull();
+  });
+
+  it("clears the latch for ordinary keys", () => {
+    expect(latchFromCapturedCombo({ modifiers: [], key: "Space" })).toBeNull();
+    expect(latchFromCapturedCombo({ modifiers: ["Alt"], key: "U" })).toBeNull();
+  });
+});
+
+// On Windows the native binary reports Right Alt as RECORD_KEY:RightAlt while
+// the focused settings window independently sees a DOM AltRight keydown. The
+// two arrive in either order, so the latch must survive both interleavings —
+// otherwise the hotkey saves as generic "Alt" and both Alt keys trigger it.
+describe("Right Alt capture on Windows (native + DOM race)", () => {
+  it("latches RightAlt when the DOM event lands first", () => {
+    let latch = nextRightModifierLatch(null, EMPTY, {
+      rightToken: "RightAlt",
+      modifierCount: 1,
+      genericModifiers: ["Alt"],
+      explicitLeft: false,
+    });
+    expect(latch).toBe("RightAlt");
+
+    latch = latchFromCapturedCombo({ modifiers: [], key: "RightAlt" });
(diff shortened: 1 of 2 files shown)


matteo8p:
I'm going to close out this GitHub issue. Please refer to the new comment I added in the GitHub issue that you posted.

PR #565, opened 2026-08-10
feat(tone): preview at the root, edit one level down

samuelalake:
`ROADMAP.md` → Freestyle Core → 3️⃣ *"Improve the UI. Simplify the tones page."*

Built and driven in the app.

## 1. Before / after

| Before: Cleanup | Before: Personal | After |
|---|---|---|
| before cleanup | before personal | after |

Before, two facets lived in one segmented control. One is cleanup, the other is how you sound
across apps. The redesign splits them into two sections.

"Cleanup" used to name a tab. The section is now "How much to fix" and the control is named
"Strength" to more accurately represent what this setting does.

The second section uses progressive disclosure and takes a preview-first approach. The root
page shows every group, the apps it covers, and what it currently sounds like. Editing happens
one level down, on the group's own page. That's the convention settings use on iOS and macOS.

---

## 2. Personal, Work, Email

| The row | The page behind it |
|---|---|
| row | personal |

Each group gets its own page holding its four options and its app list. Choosing which apps
belong to a group happens there.

The icons on the row overlap and reach a maximum size once more than five apps are added to
the group, so the row is the same width whether a group has six apps or thirty. The last app
you added shows first in the stack.

overflow

---

## 3. Everywhere else

| Before | After |
|---|---|
| before | after |

This group has no "add an app" section. You can't add an app to it. The subtitle explains
that instead. The old line said "anywhere that isn't a chat or email", which missed anything
you'd moved to another group.

---

## 4. Custom

| Before | After: the row | After: the page |
|---|---|---|
| before custom | custom row | custom page |

Custom used to open a text box that grew to fit the whole prompt, making the page 7,694px
tall with everything else below the fold.

Selecting Custom now takes you to its own page, with your last preset pre-filled the first
time. After that your own prompt is what you come back to. On the root page, Custom shows a
row with the first line of it.

Low, Medium and High each show a sample of what they do. Custom can't, since it runs
instructions only you write, so the row says what's written and takes you to where you change
it.

"Reset to presets" used to switch Strength back to Low and leave Custom altogether. It's now
**Start from a preset**: pick Low, Medium or High and that text loads into the editor as a
draft, staying in Custom. To leave Custom, use the control on the root page.

---

## What's in and out

**In:** the page structure, the app icons and grouping, the custom prompt page, and the
wording fixes above.

**Out.** These are product calls, not layout ones, so I left them alone:

- **The tone words.** The four groups use twelve words across four unrelated scales. "Casual"
  means three different things depending on the group.
- **The defaults.** Every group ships set to "Off", so the page does nothing until you visit
  it and turn something on.
- **One mislabelled option.** In Work, the setting saved as `friendly` displays as
  "Enthusiastic" and is described as "Upbeat and warm".
- **The "Cleanup is off" banner.** Tone shows this banner and points you to Models to turn
  cleanup on. For Freestyle Transcribe users, Models already has it on and greyed out, so the
  banner sends you to fix something you can't change. Separate bug.

  cleanup banner

Typecheck and Biome clean per `CONTRIBUTING.md`.


MathurAditya724:
Thank you for this, can you resolve the merge and CI issues?

samuelalake:
@MathurAditya724 the merge conflict here isn't resolvable, and I don't think this PR should be.

#570 removed the Tone page along with the rest of `pages/`, so this redesigns a surface that no longer exists. Rebasing it would mean rebuilding it against the new settings view, which is a different piece of work and probably wants a fresh look now the UI has moved.

Happy to close it. Before I do, one thing worth flagging: the tone settings keys are all still defined in `shared/settings-keys.ts` and the server still reads them in `resolveBaseCleanupPrompt`, but nothing in the new UI writes them any more.

Let me know if you'd rather I take a pass at tone in the new settings view, or just close this.

PR #567, opened 2026-08-11
feat(hotkey): show key names alongside symbols where the key is taught

### LINKED ISSUE #566: Onboarding shows the dictation key as a symbol with no name
Onboarding tells you which key to press to start dictating, but shows only a symbol. On macOS
the default key is Fn, which renders as a bare globe emoji:

hotkey badge

That's the whole hint. No word next to it, no tooltip. A first-time user who doesn't already
know the Globe key has nothing to go on, and onboarding is exactly where they don't know it
yet. The same badge appears on the Settings page afterwards, so there's no second chance to
learn the name.

## It's worse on third-party keyboards

The globe symbol assumes the key is physically printed with a globe, which is true on Apple
keyboards. On a Logitech keyboard the same key is printed **fn**. Anyone on a Mac mini, Mac
Studio or Mac Pro brings their own keyboard, so this is not a rare case. The app shows a
symbol that appears nowhere on the hardware in front of them.

The app can't detect this, and probably shouldn't try. The Swift listener identifies the key
from `CGEventFlags.maskSecondaryFn` (`native/macos-key-listener.swift`), a flag both
keyboards set identically. Getting the make and model would mean enumerating HID devices
through IOKit and then guessing, from a product name, whether a globe is printed on the key.
That means maintaining a list of Apple keyboard models and being wrong about everything else.

A label that reads correctly on both keyboards costs nothing and never goes stale.

## Where the labels come from

`keyDisplayLabel` in `use-hotkey-recorder.ts`. Modifiers resolve differently per platform:

| | macOS | Windows / Linux |
|---|---|---|
| Command | `⌘` | `Super` |
| Control | `⌃` | `Ctrl` |
| Option / Alt | `⌥` | `Alt` |
| Shift | `⇧` | `Shift` |
| Fn | `🌐` | `Fn` |

Windows and Linux users get words. Mac users get symbols. Non-modifier keys are symbols
everywhere, so Space shows as `␣` and Escape as `⎋` on every platform.

## Proposed fix

Symbol and word together, so `🌐 Fn`, `⌘ Command`, `␣ Space`. The symbol stays for anyone
who reads shortcuts that way, the word covers anyone whose keyboard is printed differently or
who hasn't learned the symbol yet. Applied where there's room: onboarding, the Settings row,
and the tutorial. Compact displays like the sidebar keep the symbol alone.

samuelalake:
Closes #566.

The dictation hotkey is shown as a symbol alone. On macOS the default is Fn, which renders as
a bare globe emoji: no word next to it, no tooltip. It appears that way in Settings and in the
recorder.

## Why a symbol isn't enough here

The globe is printed on Apple keyboards. On a Logitech the same key reads **fn**, and anyone
on a Mac mini, Mac Studio or Mac Pro brings their own keyboard. So the app names a key that
doesn't appear on their hardware.

It can't detect which keyboard you have, either. An Fn press arrives as
`CGEventFlags.maskSecondaryFn` from both. Getting make and model would mean enumerating HID
devices and then guessing whether a globe is printed on the key, which means maintaining a
list of Apple models that is wrong for everything else. Naming the key costs nothing and never
goes stale.

## The change

`keyDisplayLabel` takes a `verbose` option returning the symbol paired with its name, so
`🌐 Fn` and `⌘ Command`. `settings-view` passes it when rendering the hotkey badge.

macOS modifiers only. The table is written out in full rather than composed from symbol plus
word: composition has no good rule for the `Right ` prefix, giving `Right ⌘ Right Command`,
and `Ctrl` versus `Control` collides the same way on Windows. Space, Escape and the arrows
keep their bare symbol, since those glyphs are the convention and their names are ordinary
English that belongs in the locale files.

Threaded through `comboDisplayKeys` and `formatAcceleratorKeys`, with explicit arrows at the
`.map` call sites so the array index can't land in the new argument. There's a table test over
both platforms covering every token in the symbol maps.

Rebased onto the new UI from #570. Typecheck, Biome, knip and `vitest run` clean.

### DIFF
--- a/apps/electron/src/renderer/src/components/settings-view.tsx
+++ b/apps/electron/src/renderer/src/components/settings-view.tsx
@@ -285,7 +285,7 @@ function HotkeyRow({
           className="tavern-set-keys"
           onClick={() => recorder.startRecording()}
         >
-          {formatAcceleratorKeys(accel).map((k) => (
+          {formatAcceleratorKeys(accel, { verbose: true }).map((k) => (
             <kbd key={k} className="tavern-kbd">
               {k}
             </kbd>
--- a/apps/electron/src/renderer/src/hooks/use-hotkey-recorder.test.ts
+++ b/apps/electron/src/renderer/src/hooks/use-hotkey-recorder.test.ts
@@ -1,0 +1,82 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+/**
+ * Verbose key labels, on both platforms.
+ *
+ * Written after a review found the first attempt produced "Right ⌘ Right
+ * Command" and "Ctrl Control": it composed symbol + word and used a substring
+ * guard that missed the "Right " prefix and the Ctrl/Control mismatch. The
+ * table below is the contract, so a future refactor that reintroduces
+ * composition fails here instead of in someone's onboarding.
+ */
+async function loadWith(isMac: boolean) {
+  vi.resetModules();
+  vi.doMock("@renderer/lib/platform", () => ({
+    IS_MAC: isMac,
+    IS_WINDOWS: !isMac,
+    IS_LINUX: false,
+    PLATFORM: isMac ? "darwin" : "win32",
+  }));
+  return import("./use-hotkey-recorder");
+}
+
+beforeEach(() => {
+  vi.resetModules();
+});
+
+describe("keyDisplayLabel verbose", () => {
+  it("names macOS modifiers without duplicating the Right prefix", async () => {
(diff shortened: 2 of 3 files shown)


MathurAditya724:
Hey man 👋🏼 , thank you for this, can you resolve the merge and CI issues? so that we can merge this

samuelalake:
Rebased onto main, conflicts gone. Title fixed for the Conventional Commits check.

`Build & Test` needs a maintainer to approve the run.

Heads up: #561 touches the same two files.

PR #604, opened 2026-08-15
Remove the suggestion strip + 6/10 Wake the corner up (desktop)

matteo8p:
**Merge this next — it is the only thing between `main` and the rest of the stack.**

Contains two commits:

1. `revert(panel): remove the suggestion strip`
2. `feat(companion): fire the suggestion state, and let the bubble stand down` — this is **#597**, which merged here rather than into `main`

## Why #597 is in here

I retargeted #597's base to this branch so the stack would sit on top of the strip removal, and didn't flag clearly enough that this PR had to land first. Merging #597 therefore delivered it *here*, not to `main` — which is why `main` still shows #596 as its newest commit. Nothing is lost or out of sync; #597's work is the second commit below.

## 1. Removing the strip

The strip was pulled out of #595 before that PR merged, and #595 landed without it. It reached `main` anyway via #596: `35c4315e` — the **superseded** version of #595's commit, the one that created the strip — was still reachable from a checkout that had diverged from a force-push, and a `git pull` there merged it back in.

Removes `suggestion-strip.tsx`, its mount in `panel.tsx`, its CSS, and the `"strip"` member of `SuggestionSurface`. Zero matches for `SuggestionStrip` / `tavern-strip` across `apps/electron/src` at the tip of the stack.

## 2. Waking the corner up (was #597)

`CompanionState` has always included `"suggestion"` and `spark.tsx` renders it in full, but nothing ever sent it — `setCompanionState` had one call site toggling working/idle. A pending notification now lights that glow; the bubble collapses into it after 30s unmoused rather than sitting over the desktop until the 14-day TTL; hovering defers the collapse; and a notification arriving emits a sprite beat so the character reacts to background work.

## Stack repair

#598 had been polluted by an "Update branch" merge that pulled `35c4315e` — the strip commit — back in for a second time. I rebuilt it from its own commit and re-based #599–#601 behind it. All four are single-commit branches again and all four are mergeable.

## Testing

- `pnpm typecheck` clean (node + web)
- `pnpm test` — 80 passed
- Biome clean


matteo8p:
Superseded: both commits reached `main` when #598 merged (its base carried this branch's content). `main` has the strip removal and 6/10; nothing here is outstanding.

PR #648, opened 2026-08-19
feat(panel): dictate into todos and notes

yadgire7:
## What

Dictation had a single destination inside the panel — the chat composer. Speaking while on the Todos or Notes tab pulled you back to Chat and took the transcript with it, so the only way to add a to-do by voice was to dictate into chat and retype it.

The visible tab now decides where dictated text lands:

| Tab | Lands in | Successive utterances |
|---|---|---|
| Chat | composer | append (unchanged) |
| Todos | the add-to-do box | replace |
| Notes | the open note, or a fresh one from the list | append |

The hotkey behaves exactly as it does today — hold to talk, release to commit. Only the target changes. Dictated to-dos fill the box rather than committing outright, so a misheard item is fixable before it lands in the list.

## Why these details

- **The state machine moved out of `panel.tsx` into `lib/dictation-sink.ts`.** Partials carry the whole utterance rather than a delta, and a final replaces the partial tail instead of appending to it. That bookkeeping was inline and untested; getting it wrong duplicates the user's words. It is now pure and covered by unit tests, including replayed multi-utterance sessions.
- **Routing keys off the registered handler, not the tab alone.** With Settings open on Todos the tab is still `todos` but its field is unmounted, so routing falls back to chat rather than dropping the transcript.
- **The panel open trigger is carried through to the renderer.** Dictation opens the panel over the same `panel:focus-composer` channel as the tray and the summon hotkey. Without the trigger, suppressing that channel to protect a to-do would also stop tray and hotkey opens from surfacing the composer.
- **To-do dictation replaces rather than appends.** Appending is right for a note or a chat message that grows as you keep talking, but a to-do box holds one item — appending glued every utterance onto the first thing ever said into it.

## Testing

- 18 unit tests for the sink state machine: partial/final ordering, append vs replace, error rewind, recovery after an error, final with no preceding partial, and tab-to-sink mapping.
- Verified the tests fail when the fixes are reverted, rather than assuming coverage: reintroducing the append bug reproduces `'buy milk call the bank'`, and reintroducing the error-path bug wipes a hand-typed to-do to `''`.
- Full suite green on this base: 136 electron tests, 412 server tests, both electron typechecks, `biome check` clean across the repo.

Not yet exercised by hand in a running app — worth a manual pass on the Todos and Notes tabs.

## Behavior change worth flagging

A dictation-driven panel open no longer forces the Chat tab. Tray, summon-hotkey, and sign-in opens are unaffected and still surface the composer.

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -3504,6 +3504,12 @@ ipcMain.on("panel:set-busy", (event, busy: unknown) => {
 // Clicking into the composer after an agent tool yielded key focus: panel
 // windows don't always take key back from a content click on macOS, so the
 // renderer asks for it explicitly.
+// Transitions alone leave a recreated companion stale until the next blur.
+ipcMain.handle("panel:is-focused", () => {
+  const win = panelWindow;
+  return !!win && !win.isDestroyed() && win.isFocused();
+});
+
 ipcMain.on("panel:request-focus", (event) => {
   if (event.sender !== panelWindow?.webContents) return;
   const win = panelWindow;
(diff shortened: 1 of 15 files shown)


matteo8p:
@yadgire7 Really appreciate you putting in this PR! Taking a look.

matteo8p (changes requested):
@yadgire7 I really appreciate you putting this PR in. There's a couple of changes that I think would make it a better UX than the existing implementation. 

Right now we have 2 hotkeys, one for regular dictation and one for click-to-activate. For some reason you cannot dictate directly into the freestyle interface, which is what you're trying to solve. We use regular dictation; it temporarily hides the freestyle app and will paste behind the app. 

I think the better experience is to have the click-to-activate be left as is, but allow the user to use regular dictation to dictate into the Freestyle interface itself like the notes and the todo list. 

Let me know what you think!

Issue #182, opened 2026-06-05
Remove subtext from settings page.


matteo8p:
# Overview 

We already know what a settings page is. We don't need subtext to further back it up. Just get rid of it.

<img width="1233" height="796" alt="Image" src="https://github.com/user-attachments/assets/3f8ae531-5817-44e0-906f-8d843bb9f1e4" />

# First time contributing? 

If this is your first time contributing, I highly recommend reading our [CONTRIBUTING.md](https://github.com/freestyle-voice/freestyle/blob/main/CONTRIBUTING.md). It has instructions on how to set up. 

Also consider joining our discord server. This is where contributors communicate. We have all project discussions there. Happy to provide assistance there, too.

<p align="left">
  <a href="https://discord.gg/Fmgt5yZCDu"><img src="https://img.shields.io/badge/Discord-Join%20Server-5865F2.svg?style=for-the-badge&logo=discord&logoColor=white" alt="Discord" /></a>
</p>

matteo8p:
If you're interested in taking on this issue, please leave a comment. I will assign it to.

EvolutionX-10:
I'd like to take this, will be submitting a PR in some time.

matteo8p:
@EvolutionX-10 Sweet, thank you so much for taking this on! Assigning this issue to you. 

Please also consider joining our Discord! That's where other contributors and I hang out.

PR #654, opened 2026-08-20
fix(scheduled): reliability fixes for run-now, notifications, and thread opening

matteo8p:
Implements the desktop-side fixes from the scheduled-tasks/notifications reliability audit. **Branched on top of #652** (poll-timeout bump), so merging this covers both. Companion to freestyle-voice/cloud#182.

## Wrong-thread fixes
- **Latest-thread effect no longer clobbers explicit navigation** (`panel.tsx`): it now only seeds the initial selection (`setThread((current) => current ?? …)`), and open-thread requests are sequenced with a selection counter so a slow `getThread` can't override a newer click. This was the most likely everyday cause of reminders opening the wrong thread.
- **`panel:open-thread` rides the `PanelRendererMessageQueue`** instead of `did-finish-load`, which the queue's own docs say is not a renderer readiness signal — cold-start clicks no longer get dropped (and no longer fall back to the latest thread).
- **Local store prefers the payload's threadId over the pinned `thread_id`** and re-pins on divergence, so a cloud-side re-point (see the cloud PR) is honored instead of serving the first-ever thread forever.
- **Banner bookkeeping keyed by id + createdAt** instead of a grow-only id set: re-posted notifications banner again; `Notification` objects are retained until click/close (no GC'd click handlers); a click that resolves no threadId still opens the panel instead of silently doing nothing; nothing is marked seen when banners aren't supported.
- **Outbox never downgrades a queued `open` to a `dismiss`**; the cloud DTO type now declares `payload.threadId`.

## Run-now fixes
- **Run state lives in a module store** (`lib/run-now-store.ts`, consumed via `useSyncExternalStore`) instead of component `useState` — tab switches, hotkey summons, and thread switches (which remount the whole tab tree) no longer erase an in-flight run. Completion invalidates the task list *and* the Briefs list; per-task state means two runs can't stomp each other's status.
- **The result is finally surfaced**: a successful run shows a **View brief** button that opens the run's thread (wired `panel → BrainFiles → ScheduledTasks` via `onOpenThread`). Errors render as per-task notices.
- **A failed background refetch no longer blanks the task list** — cached data keeps rendering with a "couldn't refresh" notice; the full error card only appears when there's nothing cached. Brain-*file* load errors no longer hide the Scheduled section.
- **Hardened plumbing**: 15 s timeouts on scheduled API calls (a hung request used to pin "Running…" forever), the poll loop polls before sleeping and tolerates up to five consecutive transient failures, and a bare 409 reads "already running" instead of "unavailable".

## Testing
Component tests updated for the new gating; full electron suite (117) and server suite (412) green; typecheck and biome clean.

**Not merged, not deployed** — for review. Pairs with cloud PR #182 but degrades gracefully in either merge order.

### DIFF
--- a/apps/electron/src/main/index.ts
+++ b/apps/electron/src/main/index.ts
@@ -3173,7 +3173,8 @@ const NOTIFICATION_FALLBACK_POLL_MS = 120_000;
 let notificationFallbackPollTimer: NodeJS.Timeout | null = null;
 let stopNotificationStream: (() => void) | null = null;
 let notificationsShowing = false;
-const notifiedIds = new Set<string>();
+const notifiedIds = new Map<string, number>();
+const activeNotifications = new Map<string, Notification>();
 
 interface DesktopNotification {
   id: string;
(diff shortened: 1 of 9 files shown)


matteo8p:
Added a fix for the empty-thread-until-restart bug (commit on this branch): a notification clicked mid-run mounted `useChat` with the thread's then-empty message list, and because `PanelInner` is keyed by thread id and `useChat` treats `messages` as initial state only, every later fetch of the same thread silently discarded the fetched messages — the thread stayed empty until an app restart. The chat now adopts a freshly fetched snapshot whenever it carries more messages than the live chat and no turn is in flight. (The mid-run-notification trigger itself is closed cloud-side by #182's deferred notify.) Also merged main into this branch — conflicts with the squash-merged #652/#653 resolved.

